PR #44, opened 2020-04-11
Fixing NameError on cpmodel_path

yfukai:
Hi, I'm enjoying this wonderful package. Thanks a lot!
When I was training a model with `--pretrained_model None` as described in the README, I got the error "NameError: name 'cpmodel_path' is not defined". This commit intends to fix that problem by defining the variable.


carsen-stringer:
thanks for this!

PR #84, opened 2020-06-08
Change opencv backend package

Czaki:
because gui is created using PyQt5 then can be used open cv without qt binding. 
Using version with qt binding produces this werid messages:
```
objc[18893]: Class QMacAutoReleasePoolTracker is implemented in both /Users/gbokota/.pyenv/versions/3.6.9/envs/partseg3.6/lib/python3.6/site-packages/PyQt5/Qt/lib/QtCore.framework/Versions/5/QtCore (0x1119e9060) and /Users/gbokota/.pyenv/versions/3.6.9/envs/partseg3.6/lib/python3.6/site-packages/cv2/.dylibs/QtCore (0x14538c700). One of the two will be used. Which one is undefined
``` 
And may cause some hard to debug bugs.

See https://github.com/skvark/opencv-python/issues/341#issuecomment-640586280


carsen-stringer:
thanks so much for this! hopefully this will solve these problems for everyone (with Macs)

PR #94, opened 2020-06-22
Remove GIL lock and manual garbage collection

chrisroat:
When running this code in a worker-type application ([dask](https://dask.org/)), there were a couple performance bottlenecks. This PR tries to address these, provided these tweaks do not interfere with workloads that are typically used with this library.

### DIFF
--- a/cellpose/dynamics.py
+++ b/cellpose/dynamics.py
@@ -9,7 +9,7 @@ import mxnet.ndarray as nd
 from numba import njit, float32, int32, vectorize
 from . import utils, metrics
 
-@njit('(float64[:], int32[:], int32[:], int32, int32, int32, int32)')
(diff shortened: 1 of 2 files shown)


chrisroat:
To put a number on it, I see ~75% of time in GC toward then end of an evaluation of a 200x768x768 single-channel image.  The screenshot below is from a Google Cloud worker, showing the GC causes the iterations on the GPU to slow down as the duty cycle falls off.  At 8pm the GPU is complete and the rest of the computation proceeds (with a jump in main memory).  The jump at 8:10 is the next image being evaluated.

Screenshot from 2020-06-22 20-14-07

marius10p:
This sounds good to me, @carsen-stringer ?

PR #117, opened 2020-09-11
Fix import bug in `cellpose/plot.py` file

abh799:
Scikit Image `io` operation is used inline 91 - 94 but `io` is not imported which throws an error when running the `show_segmentation`  function with `file_name` argument


carsen-stringer:
thanks for the bug fix! I've changed this to use the cellpose `io.imread` function instead (I removed scikit-learn dependencies)

PR #139, opened 2020-10-23
Update train.rst

Xqua:
Added the fact that the path must be absolute or else it will not find the training images !


carsen-stringer:
thanks!

PR #173, opened 2020-12-27
Fixing syntax error in imagej_roi_converter.py for older Python versions.

### LINKED ISSUE #172: Syntax error for (presumably) older Jython versions
Hi,

I'm attempting to run the `imagej_roi_converter.py` script as described in the [docs](https://cellpose.readthedocs.io/en/latest/outputs.html), but I'm getting the following syntax error:

`SyntaxError: ("'with' will become a reserved keyword in Python 2.6", ('<string>', 17, 0, ''))`

I presume this has something to do with my Jython version or which Python version the installed Jython version is communicating with. I haven't been able to figure out the cause of the problem, but I have been able to fix it by slightly modifying the `imagej_roi_converter.py` script. I will be submitting a PR with my fix, in case it might be useful.

Thank you for any assistance you can provide,

V

vinisalazar:
Hi,

This is a fix for #172. if merged, this PR will:

* remove the `with` statement from the `imagej_roi_converter.py` macro;
* Add parentheses to the print statement at line 25.
* Format script with [Black](https://black.readthedocs.io/en/stable/)

I know using the `with` statement for opening files is a best practice for current Python code, however it was causing problems with older Python versions. This change fixes that.

Thank you for any assistance you can provide,

V

### DIFF
--- a/imagej_roi_converter.py
+++ b/imagej_roi_converter.py
@@ -14,16 +14,16 @@ rm = RM.getRoiManager()
 
 imp = IJ.getImage()
 
-with open(file_name, 'r') as textfile:
-	for line in textfile:
-		xy = map(int, line.rstrip().split(','))
-		X = xy[::2]
-		Y = xy[1::2]
-		imp.setRoi(PolygonRoi(X, Y, Roi.POLYGON));
-		#IJ.run(imp, "Convex Hull", "")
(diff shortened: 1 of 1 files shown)


carsen-stringer:
great, thanks!

Issue #174, opened 2020-12-27
Invert grayscale option on CLI batch mode


vinisalazar:
Hi,

I'm trying to run cellpose using the CLI batch mode using a directory of images as input. However, to get good results on my data, I must turn on the 'Invert_grayscale' option (when using the GUI). Is there a way to activate this option for the CLI application?

Thank you for any assistance you can provide,

V

PR #175, opened 2020-12-28
 Adds 'invert' arg to CLI to enable 'invert grayscale' parameter

vinisalazar:
Hi,

this is an attempted fix to #174. If merged, this PR will:

- Add the 'invert' argument to the CLI
- Call the `model.eval` function with the argument set in the invert parameter.

Thank you for any assistance you can provide,

V


carsen-stringer:
thanks for adding this!

PR #190, opened 2021-01-24
colab notebook 2D

pr4deepr:
Adding google colab notebook for 2D segmentation


its-jd:
Hi, I have trained the model using a custom dataset and now I want to use this new weights for inference.  I am loading weights while creating an instance of Cellpose class as shown in the attachment. Could you please let me know if this is the correct way to load custom weights?

image


Further, when I load weights in this way and proceed to prediction, I get the following error. 

image


Could you please let me know if the error is from the way I am loading the weights?


Thanks in advance!!

pr4deepr:
Hi, is this in google colab?? Assuming you are doing it in your local environment from what I can tell. 

Regardless, if you are using a pretrained model, check this documentation:
https://cellpose.readthedocs.io/en/latest/api.html#cellpose.models.CellposeModel

> 
> class cellpose.models.CellposeModel(gpu=False, pretrained_model=False, torch=True, diam_mean=30.0, net_avg=True, device=None, residual_on=True, style_on=True, concatenation=False)
> 
>     Parameters
> 
>             gpu (bool (optional, default False)) – whether or not to save model to GPU, will check if GPU available
> 
>             pretrained_model (str or list of strings (optional, default False)) – path to pretrained cellpose model(s), if False, no model loaded; if None, built-in ‘cyto’ model loaded
> 
>             net_avg (bool (optional, default True)) – loads the 4 built-in networks and averages them if True, loads one network if False
> 
>             diam_mean (float (optional, default 27.)) – mean ‘diameter’, 27. is built in value for ‘cyto’ model
> 
>             device (mxnet device (optional, default None)) – where model is saved (mx.gpu() or mx.cpu()), overrides gpu input, recommended if you want to use a specific GPU (e.g. mx.gpu(4))
> 

I haven't used this option in a while, but in your case it should be:
models.CellposeModel(gpu=True, pretrained_model=model_type)

Check the eval option here as well: https://cellpose.readthedocs.io/en/latest/api.html#cellpose.models.CellposeModel.eval

Your post has made me realise that I should add an option of using a custom trained model in the google colab notebook... 😃

its-jd:
Thanks for the reply. You guessed it right, I am doing everything locally.

I am creating an object of class _Cellpose()_ which combines _Cellposemodel and size model_ and while creating an object, I pass the path of the weight as model_type argument. So, in the end, the _model_type_ argument is passed to the object of the CellposeModel class. 

In the constructor of the CellPose(), I added a condition to check if the custom_weights are loaded or the default 'cyto' or 'nuclei' model.


image

Issue #220, opened 2021-03-10
add "look one level down" feature to command line


carsen-stringer:
add flag to process all subfolders of input folder

carsen-stringer:
flag added to CLI `--look_one_level_down` which will process all images in folder and subfolder

saskra:
Great enhancement! Unfortunately, the corresponding argument is not passed all the way to the executing function.

https://github.com/MouseLand/cellpose/blob/ff46c5f4ebd3e3c580be62765e38dc8108fcd041/cellpose/io.py#L148
https://github.com/MouseLand/cellpose/blob/ff46c5f4ebd3e3c580be62765e38dc8108fcd041/cellpose/__main__.py#L209

carsen-stringer:
My understanding was that this was needed for model application, not training, but I can see how it could be useful for that too. Happy to take a PR implementing this functionality.

saskra:
Your understanding is not wrong, just that I keep combining training and application in the leave-one-sample-out cross-validation. I will do a PR.

PR #254, opened 2021-05-03
Enable look_one_level_down during training as well

saskra:
This allows the look_one_level_down argument to be used even during training, e.g. for cross-validation: https://github.com/MouseLand/cellpose/issues/220#issuecomment-831144620


carsen-stringer:
looks great, thanks for doing this so quickly :) 

could you remove the ignore warnings line? I'd prefer to keep them on

saskra:
I have turned the warnings back on. (However, the same warning is then always written between the steps of the progress bar).

Issue #257, opened 2021-05-04
The new cyto2 network cannot be used from command line


thosoo:
in `__main__.py`:
```
if not (args.pretrained_model=='cyto' or args.pretrained_model=='nuclei'):
                cpmodel_path = args.pretrained_model
                if not os.path.exists(cpmodel_path):
                    logger.warning('model path does not exist, using cyto model')
                    args.pretrained_model = 'cyto'
```

When  adding `--pretrained_model cyto2` in the command line, it will obviously not work.

However, I have not tried with the full path to the models yet.

giocard:
The issue  is still not closed. I could use the cyto2 from command line by modifying https://github.com/MouseLand/cellpose/blob/4b1eaa3583f6de3918bc8df5ecdc82318c89d303/cellpose/__main__.py#L142 in the same way it has been already done for line 130.

carsen-stringer:
I'm confused, does the github version work for you? if so I'll close the issue

giocard:
No, it doesn't. When running from command line, if the diameter is set to zero the program stops at line 147 
https://github.com/MouseLand/cellpose/blob/4b1eaa3583f6de3918bc8df5ecdc82318c89d303/cellpose/__main__.py#L147
because it thinks that is a model provided by the user, and therefore it expects the diameter.
What I was trying to say in my previous post is that I had to fix line 142 by adding `or args.pretrained_model=='cyto2'`  to make it work aslo under this condition. If I set the diameter, the rpogram runs fine.

giocard:
I forgot to mention that in any case, even with a model provided by the user, the program would fail if the input diameter is set to zero, because of the following error at line 147

> UnboundLocalError: local variable 'model' referenced before assignment

carsen-stringer:
ah good catch, please try now

giocard:
I did a few tests and it seems to be fixed now. Thanks!

carsen-stringer:
thanks for your help!

PR #261, opened 2021-05-07
Fix issue #257 when model is cyto2 and diameter is zero

giocard:
It fixes the situation when no diameter is given and the pretrained model chosen is cyto2


carsen-stringer:
oops sorry just saw this, I think I fixed it

PR #284, opened 2021-07-06
Changing mask dtype to uint32 to avoid index clipping

### LINKED ISSUE #283: Force casting mask to np.uint16
Hi, thank you for this great program! 

I'm trying to segment a large image, and noticed that a mask with the index >65535 is not correctly indexed (there appears several regions with the same index). I think this is because the mask is cast to `np.uint16` around this line, though it is treated as `np.int32` beforehand.
https://github.com/MouseLand/cellpose/blob/babf2eff59265815f0462f8cbc9fb9c68e747cc3/cellpose/models.py#L593
Does it make sense to change `np.uint16` to `np.int32` or `np.uint32`?

Cheers,
YF

yfukai:
This intends to close #283 , by changing the mask `dtype` to `np.uint32`.


yfukai:
Hi, thanks @carsen-stringer for your comment! Sorry, I didn't notice this point, but I would like to check and update related codes. Yeah I think that's one option, and there is another option of warning and just writing uint16-converted png. It'll be great if I could hear your opinion on this!

yfukai:
Sorry, it took a while, but I confirmed `cv2.imwrite` cannot write `np.uint32` to PNG, while there's no problem writing TIFF by `tifffile.imsave`. Reading the code, I noticed: 
- Using GUI, the mask is assumed to be `np.uint16` [here](https://github.com/MouseLand/cellpose/blob/f80a5e3b1a1cd53678c76fce162f09c945604462/cellpose/gui/gui.py#L861) and can be saved as PNG without problems.
- When using `io.save_masks`, it [detects](https://github.com/MouseLand/cellpose/blob/f80a5e3b1a1cd53678c76fce162f09c945604462/cellpose/io.py#L319) the `dtype` and uses TIFF for `np.uint32` masks.
summing up, I think the code change is unlikely to make problems with the mask saving, but I'd be happy to hear your opinion!

elmbeech:
> 

Hi @carsen-stringer and @yfukai,

I hope you are well!
As mentioned earlier (https://github.com/MouseLand/cellpose/issues/283),  it is a huge problem for us, that cellpose non longer can generate masks with > 2^16 cells. 
We hit this limit all the time. 
With a code sequence below, which sets the mask variable dtype dynamically, the code would only use uint >  16[bit], if there were more nuclei segmented than that.
If then someone gets the png related error, then the error is justified because cellpose segmented more than 2**16 nuclei.
@carsen-stringer, would this be an acceptable solution?
Should I write a pull request?

```
masks = np.zeros((nimg, shape[1], shape[2]), np.uint32)
p = np.zeros(dP.shape, np.uint32)

[...]

if (mask.max() < 2**8):
    dtype = np.uint8
elif (mask.max() < 2**16):
    dtype = np.uint16
else:
    dtype = np.uint32
mask = mask.astype(dtype)
```

Thank you, Elmar

elmbeech:
I just submitted the pull request #411 .
Hope this will do. 
Best, Elmar

chrisroat:
My preference is using int32 or uint32, which are fairly standard.  My library that uses cellpose converts the cellpose output to int32 to maintain compatibility with scikit-image segmentation does (e.g. skimage.segmentation.watershed).

Saving out to formats that only support 16-bit could generate an error/warning if pixels are out of range, otherwise it can just convert from the 32-bit.

elmbeech:
As far as I understood, outputting masks always as uint32  is what @yfukai did, and @carsen-stringer not accepted.

@chrisroat, are you sure the skimage segmentation output is in int32? Does this segmentation generate negative labels?
If you have only positive values, why would you convert uin32 into int32? 
You risk clipping your labels by  65'535 (which is actually the same as saving  files as uint16) and no warning will be spit out!
With uint32 Your max value is 4'294'967'296  (same as int64)!

I just would like to have a cellpose that can segment more than 65'535 nuclei.
So, what do you want me to do to resolve this issue?

Thank you, Elmar

chrisroat:
> Hi @chrisroat,
> 
> As far as I understood, outputting masks always as uint32 is what @yfukai did, and @carsen-stringer not accepted.
> 
> And @chrisroat, are you sure the skimage segmentation output is in int32? 

I guess I wrote it because that is what I get when using the watershed.  Do you have a counter-example?

> Does this segmentation generate negative labels? If you have only positive values, why would you convert uin32 into int32? You risk clipping your labels by 65'535 (which is actually the same as saving masks as uint16) and no warning will be spit out! With uint32 Your max value is 4'294'967'295 (same as int64)!

That is not correct.  The ~4e9 number is the number of values available for BOTH uint32 and int32.  (They are both 32 bits -- just in once case, half the numbers are negative.)  int64 has ~1e19 values (2^64 = 2^32 * 2^32) which is way more than most people need.

In the general case of image analysis, negative labels are sometimes used to indicate background, which might also be segmented.  

Yes, uint will contain twice as many positive values, but it's really not super important.  Generally, I avoid uints unless I really need them; too much pain when I run into things like one minus two is a big positive integer.

> I just would like to have a cellpose that can segment more than 65'535 nuclei. So, what do you want me to do to resolve this issue?

If someone has created a branch usint int32/uint32, then you can likely use it yourself.

elmbeech:
oh yes, you are right! my bad. my math was very wrong! sorry!

yfukai:
Just resolved the conflict! Any comments are welcome.

elmbeech:
thank you @yfukai !
some strange error.  that pandas or dask is not installed.
anyhow, I hope this branch get merged into master.

yfukai:
Thanks @elmbeech, still not sure where the error comes from, but I'm happy to fix if necessary.

carsen-stringer:
thank you all for your work on this, cellpose should work now with many cells!

yfukai:
Thanks @carsen-stringer, I'll check how it works!

elmbeech:
Thank you, @yfukai  and @carsen-stringer!
This is super cool! I will test it too. Happy that cellpose again can handle meany cells.

PR #300, opened 2021-07-17
Skeleton-Cellpose 

kevinjohncutler:
These changes enable cellpose to work well on arbitrarily long and anisotropic cell morphologies. This implementation utilizes 4 prediction classes, so it is important that users delete existing saved flow fields before training. The 'skel' flag (for morphological skeleton, the basis of earlier implementations) defaults to True but can be set to False to obtain the old cellpose behavior. If nclasses = 3 is specified, then older models are compatible with the new code. Old center-seeking flow models used with skel=True work well (much less over-segmentation). 

Bug fixes include the gpu flow computation (there was 'bleedout' in the diffusion simulation) and an issue where single-channel images were transposed independent of the labels, causing training to fail. 

Many more changes included in this code will be detailed in an upcoming paper.


cellpose/transforms.py
```diff
@@ -212,15 +212,18 @@ def move_axis(img, m_axis=-1, first=True):
     img = img.transpose(tuple(axes))
     return img
 
+# This was edited to fix a bug where single-channel images of shape (y,x) would be 
+# transposed to (x,y) if x<y, making the labels no longer correspond to the data.
```
sommerc, line 216: Thanks for finding this 💯, I was going crazy already :-)

PR #333, opened 2021-09-23
Fix "local variable 'nimg' referenced before assignment"

oeway:


cellpose/transforms.py
```diff
@@ -646,6 +646,8 @@ def random_rotate_and_resize(X, Y=None, scale_range=1., gamma_range=0.5, xy = (2
             amount each image was resized by
 
     """
+    nimg = len(X)
+
     if inds is None: # only relevant when debugging 
         inds = np.arange(nimg)
```
oeway, line 652: If inds is None, nimg becomes referenced before assignment

carsen-stringer:
I think this is fixed now, thanks!

PR #341, opened 2021-10-01
Allow empty masks during training

erjel:
Hi,

In one of my projects I see a large number of false positive detections in certain areas of a larger where I do not expect any cells. The large number of FP arised since I only included image patches into the training data which where densely packed with cells. 

To balance the training data, I included regions with no cell labels. Unfortunately cellpose could not handle the empty masks. I created corresponding test cases and fixed the problematic parts in the code.

~~Note that I only test the gpu verison of `mask_to_flows` (since I am currently not interested in the CPU version). If no GPU is detected, the tests are skipped. That said, it is probably good practice to check the CPU version for similar problems as well.~~

Eric


carsen-stringer:
thanks, this is good to add. but alternatively, rather than changing the n_iter in the dynamics flow function, can you set it so that the function returns zeros at the beginning if there are no masks? that way it is an easy modification for both `_gpu` and non gpu version

erjel:
Hi @carsen-stringer ,

sorry was afk for a while. I changed the code as suggested and modified the  CUDA checks in `core.py` (since they prevented testing on linux systems without NVIDIA drivers).

Eric

erjel:
Note that 05d10855dd8892d98f0a875326782d6d81eb3c29 
fixes #374
and also fixes #378, fixes #379, fixes #401, fixes #406 which are all duplicates of the already mentioned issue.

Jooppie:
Hi, is this added to version 1.0. I get errors with empty masks  even with --min_train_masks 5 . I prefer using the empty masks for the same reason as Erjel.

carsen-stringer:
this should be working now thanks for all your work tracking down these bugs

PR #356, opened 2021-10-28
Add distributed segmentation library

### LINKED ISSUE #244: Processing Large 3D image
Hi @carsen-stringer ,
In an existing [issue thread](https://github.com/MouseLand/cellpose/issues/178) I found that you recommended :  _**“if you're running out of memory due to the size of the tiff we don't currently have a workaround for that -- we recommend splitting the stack and stitching the masks afterwards.”**_

I’m wondering how to achieve an optimal stitching of the masks (aka Label images, in which all the pixels of an object have the same value). Indeed stitching of overlapping intensity values looks natural but my first guess is that it is much more complicated for Labels, in 3D! Detecting objects in 3D in the overlapping area, detecting partially/fully overlapping objects, deciding which one to keep … removing objects touching the edge ?
Am I overthinking it? 

With @lacan we thought it might be “easier” to fuse the flow images ? maybe in temporary files to overcome memory issues and run the other section of CellPose after on the fused image, which might not be so memory hungry?
Best,
Romain

chrisroat:
Added a `contrib` directory for contributed utilities.  Used another `extras_require` in `setup.py` to handle additional dependencies.

Closes #244


carsen-stringer:
Hey Chris, the github workflow is running out of memory when trying to run your volume: https://github.com/MouseLand/cellpose/runs/4132869767?check_suite_focus=true. I'm going to comment out the tests for now but let me know if you want to use a smaller volume or different settings for the tests

chrisroat:
Oh.  We should have tested this before merging.  You may now have a big useless testdata file in your git history.  :(

Do you have guidance on how big a file could be used?  I could crop the current data smaller -- but then I worry about the overlap regions getting different segmentations when each block is segmented individually. 

The testdata I had was 80x448x224 that I snipped from some data I had around.   It was designed to be two adjoining 224x224 chunks which the could be stitched together.   My understanding was that chunks should be at least 224x224 so that the overlapping area (I used ~30 pixels depth) would get approximately the same segmentation.

GFleishman:
While developing a parallel solution, I did the exact same thing (test data had only two blocks). I always use an overlap of 2x the given cell diameter, which for my data is 25-30 pixels; but I can see how this would be problematic for smaller cell diameters. Probably the ideal overlap would be the radius of the full model FOV, which we could figure out from the model layers and kernel sizes, but I didn't bother since my stitching results seemed to look ok.

But I'm confused why 224 is necessary?

chrisroat:
The solutions on each side of a chunk gap must approximately match.  I
found that the cellpose model wasn't very good at giving the same result in
overlap regions unless the overlap was about that size -- I think I got 224
from an early understanding of how tiling/rotating was working within
cellpose.

C

GFleishman:
Thanks for the info!

I'm still a little confused though - are you using a depth of 15 (which would produce an overlap region of 30; right block goes into left block territory by 15 and vice versa, so 30 total pixels are "segmented twice") or a depth of 30 (for an overlap of 60)? My depth is about 60 (2x cell diameter, which is like 26-30), so my overlap region ends up being ~120. The two sides are pretty consistent, but one problem still emerges at the stitching step.

Consider two abutting block faces. Say for the left block, you segmented a region into two adjacent cells. Say for the right block you got *almost* the same result, but the boundary between the two cells is different from the boundary in the left tile by just one pixel. If you naively merge using the label module structuring element strategy, all four regions (which really should be two cells) get merged into one thing.

I hacked an imperfect correction for this here: https://github.com/GFleishman/cellpose/blob/33edb3967558e5d832fafac7f1969f0ba55a57b0/cellpose/distribute.py#L258

I didn't look at any tiling/rotating aspects of the data in cellpose - did I miss something important? I figured as long as the overlap region was large enough that neighboring tile results were consistent within correction limits along the boundaries, that the block could be whatever size you want - e.g. (crazy but) you could do a 1x1x1 block with overlap size of 60 - so 121x121x121 going into cellpose.

chrisroat:
If you aren't getting the same result on both sides on the boundary pixels,
it likely means your overlap isn't big enough.  You really need the same
result on both sides.  I'd suggest putting together a mvce test case for
your problem, rather than describing in words.

If the cellpose developers could provide guidance on how big an area is
needed to assure the same result for a pixel, it would be helpful.
Otherwise, we just have to keep increasing it when we see an inconsistency.

Also, don't expect perfection.  As long as the error rate (over or under
segmentation) on the boundary is below the error for the algorithm in the
bulk, there isn't much point in improving it.

carsen-stringer:
given the downsampling is a factor of 8 and the convs are size 3, you can imagine strongest effects in an area of around 24 pixels, but there are 4 conv layers per block so the effects can be even larger. I would suggest at least as big as a couple cell diameters but have not tested these overlap issues as extensively as you both have. cellpose does tile the image in 224x224 sections if it's >224 in size. if it's less, then it pads to make it a multiple of 16 and runs on that.

@chrisroat regarding the test volume, could it be fewer planes in Z?

chrisroat:
I see no reason why not.  The depth isn't important, other than to make sure cellpose will work effectively.  What z depth do you think is good in this dataset?

Alternatively, feel free to change the testdata out for your own dataset.  It just should have enough xy extent to have 2 chunks such that the chunks give the same results in the dask-overlap depth.

Now that you've removed the testdata from the repo, you are likely better positioned to make the change.  [I'm curious for the reason the files were deleted.]

Some things to note:
- The test was changed to chunk by 128 instead of 224.  That should be changed back.
- The algorithm has changed a bit, so the test will currently fail because a small fraction of pixels of changed.  The expected output for the test will need to be updated.

chrisroat:
What have you tried or looked at so far?  The [`segment` function comments](https://github.com/MouseLand/cellpose/blob/5fb74fde155c35dafcd096fd66118b3cfdd174a0/cellpose/contrib/distributed_segmentation.py#L30-L67) should get you started.  You'll just need to install with ``pip install cellpose[distributed]`` to get the dependencies.  

I think the tests were commented out without any replacement, so it's possible the core code has drifted and this may not work.  Give it a go and if you run into problems, I might be able to find some time to help.

Eddymorphling:
> What have you tried or looked at so far? The [`segment` function comments](https://github.com/MouseLand/cellpose/blob/5fb74fde155c35dafcd096fd66118b3cfdd174a0/cellpose/contrib/distributed_segmentation.py#L30-L67) should get you started. You'll just need to install with `pip install cellpose[distributed]` to get the dependencies.
> 
> I think the tests were commented out without any replacement, so it's possible the core code has drifted and this may not work. Give it a go and if you run into problems, I might be able to find some time to help.

Sorry for the delay. Just had the chance to get back to this now. How would I define my `input `in `segment`? I do have `zarr `datasets that I would like to analyse using distributed segmentation. I have `cellpose[distributed]` already setup.

chrisroat:
I'm happy to help, but am away from keyboard until Jun 17.  If the code hasn't rotted too badly, I'm happy to take a trip down memory lane then.

In the meantime, you can try taking a look at the tests.  You'll find it easier to start with small examples (even if they seem trivial) before working up to your big example.

Eddymorphling:
@chrisroat Thank you. That would be great! I will send a reminder as we get closer to Jun 17.

chrisroat:
It looks like the tests & then the code were deleted over the years.  It will take more time than I have to get this working again.  Sorry!

Eddymorphling:
Thanks for your time! Happy to try it out if you would like me to beta test anything. Please let me know.

PR #395, opened 2021-12-05
WIP: Enable (local) GPU testing

erjel:
Hi,

during my last development sessions on the code, I realized that the GPU implementation is a blind spot in the test routines. While github-hosted runners do not come with enabled GPU hardware, I believe that many developers/ users have access to GPUs (at least smaller/older ones which are perfectly fine for testing purposes). By providing the necessary configuration files, everyone could check their pull requests locally for bugs in the GPU code prior to a commit.

I already started with a corresponding tox configuration file for a previous pull request, and would like to share the work and to start the discussion.

My current setup:
- tested Windows 10 and CentOS 7
- [Miniconda ](https://docs.conda.io/en/latest/miniconda.html) non-admin installation with default settings
- Install [tox-conda](https://github.com/tox-dev/tox-conda) installation (for pytorch support) in base environment for testing
- Use modified tox configuration (`tox.ini` vs  `tox-conda.ini`) for testing the code on a GPU
```
conda activate base
pip install tox-conda
tox -c tox-conda.ini
```

Next steps:
- [ ] Write/ modify a test case to check GPU utilization
- [ ] adapt miniconda/tox-conda in the github action yml
- [ ] Make sure that github actions/ tests run fine even without available GPUs

Since github actions and tox/tox-conda are rather new tools for me, I would like to ask for additional opinions on this before I invest more time in this. Currently, I see the benefits of (local) GPU testing, but I might have missed a point. What is your opinion?

Eric


carsen-stringer:
thanks @erjel indeed this would be very useful! I agree that not having GPU tests is not great, and being able to run them locally is certainly useful enough (doesn't need to be integrated into github actions). is `tox-conda` necessary for this if it's only for local testing? as a developer can I simply run `pytest tests/` the same way as usual?

Issue #380, opened 2021-11-21
Index error when training Omnipose in colab


XinChen1020:
I want to training a Omnipose based on my own dataset on colab but it keeps giving me index error

2021-11-21 17:14:46,035 [INFO] WRITING LOG OUTPUT TO /root/.cellpose/run.log
2021-11-21 17:14:48,953 [INFO] ** TORCH CUDA version installed and working. **
2021-11-21 17:14:48,953 [INFO] >>>> using GPU
2021-11-21 17:14:48,953 [INFO] >>>> Omnipose enabled. See https://raw.githubusercontent.com/MouseLand/cellpose/master/cellpose/omnipose/license.txt for licensing details.
2021-11-21 17:14:49,097 [INFO] >>>> training from scratch
2021-11-21 17:14:49,097 [INFO] >>>> during training rescaling images to fixed diameter of 30.0 pixels
2021-11-21 17:14:49,174 [INFO] Training with rescale = 1.00
2021-11-21 17:14:49,346 [INFO] train channels = 2
2021-11-21 17:14:49,363 [INFO] test channels = 2
2021-11-21 17:14:49,363 [INFO] flows precomputed
2021-11-21 17:14:49,404 [INFO] flows precomputed
2021-11-21 17:14:49,409 [INFO] >>> Using RAdam optimizer
2021-11-21 17:14:49,547 [INFO] >>>> median diameter set to = 30
2021-11-21 17:14:49,547 [INFO] >>>> training network with 2 channel input <<<<
2021-11-21 17:14:49,547 [INFO] >>>> saving every 100 epochs
2021-11-21 17:14:49,547 [INFO] >>>> LR: 0.00020, batch_size: 2, weight_decay: 0.00001
2021-11-21 17:14:49,547 [INFO] >>>> ntrain = 58
2021-11-21 17:14:49,547 [INFO] >>>> rescale is 1
2021-11-21 17:14:49,547 [INFO] >>>> ntest = 6
2021-11-21 17:14:49,547 [INFO] (2, 258, 252)
Traceback (most recent call last):
  File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.7/dist-packages/cellpose/__main__.py", line 436, in <module>
    main()
[...]

PR #399, opened 2021-12-08
Fix for issue #380

pascal-pfeiffer:
* omni arg was missing for tests label flow generation


pascal-pfeiffer:
Fix for issue #380

authman:
Nice catch Ilu. Maybe consider also adding in the other missing params as well `use_gpu` and `device` in addition to `omni`.

CozyDoomer:
@pascal-pfeiffer are you still working on this?
I could add the 2 missing parameters as well since I also ran into this bug.

Would make sense to me to merge this soon since I've seen it pop up everytime someone tried to train omnipose models with a test set

pascal-pfeiffer:
Added the other two params.
Should be good to merge.

carsen-stringer:
thanks this has been fixed

PR #411, opened 2021-12-18
set the masks uint variable type dynamically

elmbeech:
Hi there! 

This is the second try to resolve issue #283 pull request #284 .
I introduced some dynamical variable typing to set the masks uint type.

Elmar

### DIFF
--- a/cellpose/models.py
+++ b/cellpose/models.py
@@ -718,7 +718,6 @@ class CellposeModel(UnetModel):
             else:
                 masks = np.zeros((nimg, shape[1], shape[2]), np.uint32)
                 p = np.zeros(dP.shape, np.uint32)
-                nlabel = 0
 
                 tr = [[]]*nimg # trace may not work correctly with multiple images currently, still need to test it 
                 resize = [shape[1], shape[2]] if not resample else None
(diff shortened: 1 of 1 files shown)


yfukai:
Thank you @elmbeech for this pull request! I could not take time to resolve conflicts in my original issue after major update, and I think your new attempt is helpful.
I think one potential problem with dynamical `dtype` would be the computational cost of `masks.max() ` itself. I'd rather stick to static `dtype`, but for dynamical `dtype` options, I think there should be an option to disable this feature. Maybe we can ask for an opinion of @carsen-stringer on this...

elmbeech:
Hi @yfukai, I not at all insist on that dynamical variable typing implementation. 
If your changes were good enough, but were just conflicting with the major update, then I can simply re-submit your changes.
The only thing I aim for is, that the 32uint issue is resolved in the main branch.
Whatever @carsen-stringer think is need, I can try to do.  Elmar

carsen-stringer:
unfortunately we use cv2 for resizing (does not take uint32) so this change breaks more downstream code. I've modified the typing inside the compute_masks function to take care of that now and tested with a 9000x9000 image with 80000 cells. check out the master branch to try it

elmbeech:
Thank you, @carsen-stringer for taking care of the issue and solved the problem!
Thrilled that cellpose compute_masks now can handle more than 2**16 cells.

PR #416, opened 2021-12-31
Remove global logging setup

### LINKED ISSUE #295: turn off logging configuration
Upon import, cellpose takes over the logging apparatus, changing the logging format, logging streams, and logging level.  This is non-standard practice, and could cause problems for users (say, they environment has no disk or has precious disk space... or the cellpose logging level is way too verbose in other parts of the system).

Would it be acceptable to move this configuration to the main of the cellpose scripts, and leave the main logging config untouched when cellpose is imported as a library?

chrisroat:
As per #295, libraries generally avoid setting up logging.  That should left to the applications calling the code.  

The default logging writes to a file in the home directory, which leads to trouble if multiple runs of cellpose are executing which access the same home directory, or the program does not have write access to that directory.

Closes #295


timblosser:
I don't have much to add, except moral support. I hope this PR can be merged to master soon as I would greatly appreciate implementing this change in my workflow (MERFISH analysis).

carsen-stringer:
this is great thanks, I am happy with the new implementation, I will call the logger in the jupyter-notebook example as well to activate it for the user.

pskeshu:
is it possible to release it on pip?

PR #434, opened 2022-02-01
Update __main__.py

lacan:
Error in omni log text that was throwing an error...


carsen-stringer:
thanks this has been fixed

PR #437, opened 2022-02-06
iou patch

acycliq:
I think it makes more sense to me in cases where you have a 3d image and let's say there is a mask on some page of the z-stack that completely overshadows another mask on the next page (ie it is not just a partial overlap) the stitching algorithm should consider these two masks as belonging to the same cell regardless what the IoU statistic says

For example, if you have a case like  the one shown below, the mask at the (almost) top left of page 0 occupies the same area that the top-left mask of page 1 does **and** a lot more. The mask when you move from page 0 to page 1 shrinks and keeps its center almost at the same point. In this case, I think these two masks should be labelled the same whatever the formal IoU value is.

two_masks

This PR is a small patch to handle these cases

Some code with the expected results is included below
```
import pandas as pd
import numpy as np
from cellpose.utils import stitch3D

mask_0 = pd.read_csv('https://raw.githubusercontent.com/acycliq/Anne-s_viewer/master/temp_data/mask_0.csv').values
mask_1 = pd.read_csv('https://raw.githubusercontent.com/acycliq/Anne-s_viewer/master/temp_data/mask_1.csv').values

masks = np.stack([mask_0, mask_1]).astype(np.uint16)

out = stitch3D(masks, stitch_threshold=0.5)

print('Before stitching, mask_0 at (x, y) = (18, 17) has label %d' % mask_0[17, 18])
print('Before stitching, mask_1 at (x, y) = (18, 17) has label %d \n' % mask_1[17, 18])

print('After stitching, mask_0 at (x, y) = (18, 17) has label %d' % out[0, 17, 18])
print('After stitching, mask_1 at (x, y) = (18, 17) has label %d \n' % out[1, 17, 18])
```

Output:
```
Before stitching, mask_0 at (x, y) = (18, 17) has label 4968
Before stitching, mask_1 at (x, y) = (18, 17) has label 5221 

After stitching, mask_0 at (x, y) = (18, 17) has label 4968
After stitching, mask_1 at (x, y) = (18, 17) has label 4968 
```

Thanks!

### DIFF
--- a/cellpose/metrics.py
+++ b/cellpose/metrics.py
@@ -181,6 +181,17 @@ def _intersection_over_union(masks_true, masks_pred):
 
     iou: ND-array, float
         matrix of IOU pairs of size [x.max()+1, y.max()+1]
+
+    to_fwd: coo_matrix
+        matrix of the pairs of labels from masks_true that are fully covered
+        by a mask from masks_pred. For example (to_fwd.row[i], to_fwd_col[i]to_fwd_col[i])
+        means that the mask labeled as to_fwd.row[i] (on the next page of the z-stack) is
+        totally covered by another mask which is labeled as to_fwd.col[i] (on the current page
+        of the z-stack)
+
+    to_bwd: coo_matrix
+        Same as to_fwd but now it looks for masks in masks_pred that are totally covered by
+        a mask in masks_true
     
     ------------
     How it works:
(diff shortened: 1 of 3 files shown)


carsen-stringer:
I have mixed feelings adding this as it might make sense in some anisotropic cases to call those 2 different cells. if you decrease the stitch_threshold that would solve most of your problems because cellpose only uses the best match for stitching

Issue #464, opened 2022-03-07
stitch_threshold does not work anymore


saskra:
Some recent update seems to have broken this feature, it used to work. (But maybe I used it before only from Python scripts and not from the terminal). No matter what value I enter, I always get the same unstitched result. Maybe I can create a pull request for this.

```shell
python -m cellpose --dir test/data3D/ --img_filter _img --diameter 30. --save_tif --verbose --pretrained_model models_all_eoe/cellpose_residual_on_style_on_concatenation_off_allScratch_2022_01_27_14_48_28.419259 --savedir test/cp_results/ --use_gpu --z_axis 0 --stitch_threshold 0.1
2022-03-07 12:08:01,559 [INFO] WRITING LOG OUTPUT TO /home/saskra/.cellpose/run.log
2022-03-07 12:08:07,871 [INFO] ** TORCH CUDA version installed and working. **
2022-03-07 12:08:07,872 [INFO] >>>> using GPU
2022-03-07 12:08:07,872 [INFO] >>> running model models_all_eoe/cellpose_residual_on_style_on_concatenation_off_allScratch_2022_01_27_14_48_28.419259
2022-03-07 12:08:07,875 [INFO] >>>> running cellpose on 1 images using chan_to_seg GRAY and chan (opt) NONE
2022-03-07 12:08:08,068 [INFO] >>>> using diameter 30.00 for all images
2022-03-07 12:08:08,070 [INFO] 0%|          | 0/1 [00:00<?, ?it/s]
2022-03-07 12:08:08,075 [INFO] Evaluating with flow_threshold 0.40, mask_threshold 0.00
2022-03-07 12:08:08,076 [INFO] multi-stack tiff read in as having 21 planes 1 channels
[...]
2022-03-07 12:08:10,580 [INFO] 100%|##########| 21/21 [00:02<00:00,  8.48it/s]
2022-03-07 12:08:10,580 [INFO] network run in 2.48s
2022-03-07 12:08:10,580 [INFO] mask_threshold is 0.000000
[....]
2022-03-07 12:08:14,740 [INFO] mask_threshold is 0.000000
2022-03-07 12:08:14,975 [INFO] masks created in 4.39s
2022-03-07 12:08:15,811 [INFO] 100%|##########| 1/1 [00:07<00:00,  7.74s/it]
2022-03-07 12:08:15,811 [INFO] >>>> completed in 7.939 sec
```

saskra:
Looks like this has been resolved with the new version.

PR #465, opened 2022-03-07
Repair stitching

saskra:
Simple solution to this problem: https://github.com/MouseLand/cellpose/issues/464


simonlsp:
Thanks, this fix worked for me.

saskra:
Looks like this has been resolved with the new version.

PR #494, opened 2022-04-11
Optimize GPU memory usage

chaichontat:
Explicitly remove tensors after use to prevent out of memory errors in systems with lower GPU memory.


carsen-stringer:
sorry can you please reopen this pull request with the main branch (master was renamed)

PR #528, opened 2022-06-21
Fix bug with single object masks

erjel:
During one of my GPU model trainings I ran into a problem in `cellpose/dynamics.py`. The line: 
https://github.com/MouseLand/cellpose/blob/e6d7d2e73fe27754288de1fb30ea71bf295930e9/cellpose/dynamics.py#L88
fails if a label image contains only a single object since `squeeze` operates on all singular array dimensions. Here it should operate only on the first dimension:
```
    mu_torch = np.stack((dy.cpu().squeeze(0), dx.cpu().squeeze(0)), axis=-2)
```
To verify the bug (and test the fix), I adapted the `tox` test test runs such that it now can run locally on a GPU via:
```
tox -e gpu
```
This only requires an installed NVIDIA GPU and the its drivers. CUDA and pytorch are installed via `tox-conda`. This btw closes #395 as well.

I also noticed that large fractions of the library installations are not used during the `pytest` runs and therefore deleted them from the github actions. The GPU test is skipped during the github action runs and the cpu-only version of pytorch is used.

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -25,38 +25,20 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
+        uses: conda-incubator/setup-miniconda@v2
         with:
+          auto-activate-base: true
           python-version: ${{ matrix.python-version }}
-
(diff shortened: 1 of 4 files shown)


carsen-stringer:
thanks @erjel ! it looks like your newly added test was skipped, is that because cuda isn't available in github actions / installed correctly?

erjel:
Yes, it is skipped since pytorch couldn't find CUDA on the GitHub action machine.

PR #597, opened 2022-11-08
Checking modifier keys using bitfield operations

mark-dawn:
Hi, thanks for the great software.
I found a small bug in the GUI, where the `PyQt5.QtCore.Qt.KeyboardModifiers` objects of key and mouse events were checked using the equality operator, which fails when more than one modifier is held or the system reports extra modifiers as active.
There was also a small section of dead code: https://github.com/MouseLand/cellpose/blob/377cb69ec101ffb8d67197881290f07abeda213a/cellpose/gui/gui.py#L795-L799
that could never be reached as of the condition: https://github.com/MouseLand/cellpose/blob/377cb69ec101ffb8d67197881290f07abeda213a/cellpose/gui/gui.py#L701-L703
Those shortcuts are already set in `gui/menus.py` so I also excised those lines to avoid confusion.

### DIFF
--- a/cellpose/gui/gui.py
+++ b/cellpose/gui/gui.py
@@ -698,9 +698,7 @@ class MainW(QMainWindow):
     def keyPressEvent(self, event):
         if self.loaded:
             #self.p0.setMouseEnabled(x=True, y=True)
-            if (event.modifiers() != QtCore.Qt.ControlModifier and
-                event.modifiers() != QtCore.Qt.ShiftModifier and
-                event.modifiers() != QtCore.Qt.AltModifier) and not self.in_stroke:
+            if not (event.modifiers() & (QtCore.Qt.ControlModifier | QtCore.Qt.ShiftModifier | QtCore.Qt.AltModifier) or self.in_stroke):
                 updated = False
                 if len(self.current_point_set) > 0:
                     if event.key() == QtCore.Qt.Key_Return:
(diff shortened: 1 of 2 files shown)


carsen-stringer:
thanks, what is the advantage here? I think the user should only have one modifier clicked at a time, otherwise the program doesn't know what the user wants to do

mark-dawn:
Sorry I wasn't clear in my explanation, I noticed that the GUI was not working properly on my pc (none of the modifier shortcuts worked), and by debugging I found that qt is weird and sets some extra fields in the key modifier bits on my desktop environment, resulting in the event modifier never being strictly equal to just Control or Alt, leading to the gui just never executing those action.

So those changes have two advantages:

1. working on my machine (and possibly some other desktops) :grimacing: 
2. it **does not change old behavior** when a _single_ modifier key is pressed, so it could be merged without affecting the experience of people who don't have my problem. However the changes will slightly change what happens if the user presses multiple modifiers; as you have pointed out the user shouldn't be pressing more than one, but I will list them so you can decide if I should implement extra guards to restore the old edge-case behavior while keeping my fix working:

   - since most checks in [gui.py](https://github.com/MouseLand/cellpose/pull/597/commits/fea492a16d994cfe34f3565220f3ef9bb4d4a3ac#diff-5333dbc8af9da7949c2da18968ba665152591b5f5e1e139c494101da73ba3e63) are expressed as unions of negatives (`ev != mod1 and ev != mod2 and ....`), before my change holding a single modifier would stop the user from using the toggle shortcuts or the Z-shortcuts, but holding more than one would restore their function, now instead they will be blocked if at lest one modifier is pressed;
   - meanwhile in [guiparts.py](https://github.com/MouseLand/cellpose/pull/597/commits/fea492a16d994cfe34f3565220f3ef9bb4d4a3ac#diff-01c6ee697036f1debd16cced19afc08b1d3ac0b983169ecaba18d5ec5b68e5f6), before my change, holding more than one modifier button would stop any CTRL/ALT/SHIFT modifier from working, now SHIFT will work even if other modifiers are pressed (and still only when in drawing mode of course), same for ALT and CTRL, with CTRL taking precedence over ALT.

I am pretty sure no one relied on any of the old behaviors as they are particularly confusing or just lead to no functions, so I just ignored those weird case when fixing my issue.

PR #615, opened 2022-12-02
Fixed model loading behavior when using GPU device other than cuda:0

jasonkena:
Currently, `load_state_dict` assumes that the device used is `cuda:0`, leaving the CellposeModel randomly initialized if the target device is different than `cuda:0`. This fix replaces the `cpu` argument of `load_model` with `device`, to allow proper checkpoint reloading.


carsen-stringer:
thanks I didn't realize that pytorch wouldn't use the GPU that the model is put on (here: https://github.com/MouseLand/cellpose/blob/377cb69ec101ffb8d67197881290f07abeda213a/cellpose/core.py#L142)

PR #625, opened 2022-12-10
Mitigated FP-precision problem in label to flow conversion code and double input file detection bug

koerner-axs:
Replaced imprecise calculations of the form log(1+x) with specialized functions from Torch and NumPy.

The previous implementation caused numerical instability when very long cells are converted to flows.
I observe the diffusion process implementation to be quite robust, however going from mass concentration to log mass concentration + 1 is introducing instability. This is because to potentially very small mass concentrations (maybe 10e-50 in my particular case) we add 1.0 and only then take the logarithm. Even for float64 this gap between the exponents is too large.

To resolve this problem, a number of libraries provide specialized functions to perform the log(1+x) operation using a single call to a function commonly named log1p.

In general, I am skeptical about the need to compute the log mass concentrations. Because the flow field is normalized afterwards anyways, dropping the logarithm seemed to work fine for me.

### DIFF
--- a/cellpose/dynamics.py
+++ b/cellpose/dynamics.py
@@ -78,7 +78,7 @@ def _extend_centers_gpu(neighbors, centers, isneighbor, Ly, Lx, n_iter=200, devi
         Tneigh *= isneigh
         T[:, pt[0,:,0], pt[0,:,1]] = Tneigh.mean(axis=1)
     del meds, isneigh, Tneigh
-    T = torch.log(1.+ T)
+    T = torch.log1p(T)
     # gradient positions
     grads = T[:, pt[[2,1,4,3],:,0], pt[[2,1,4,3],:,1]]
     del pt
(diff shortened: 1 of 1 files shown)


koerner-axs:
This image shows the dFdy, that it computed for one of my labeled training images. In the top right, you can see what issues I am trying to circumvent with this pull req.
example-precision-error

koerner-axs:
Last commit also takes care of this issue: https://github.com/MouseLand/cellpose/issues/626

carsen-stringer:
thanks can you remove the second fix? I will put testing the log1p on my to-do list (want to test the full training with the original cellpose training set)

koerner-axs:
I have now reset my fork to the log1p commit.
I really should have split these two issues up.

carsen-stringer:
thanks @koerner-axs we tried using the logexp function in pytorch, but found removing the log computation altogether worked best. this will be available in the new release

PR #634, opened 2023-01-09
Erjel main

carsen-stringer:
changed it so that pytorch by default would install via pip (since that is the default install)


carsen-stringer:
thanks @erjel!

Issue #581, opened 2022-09-28
Export of native ImageJ .roi files via roifile


Etherian:
Importing Cellpose outlines into ImageJ currently requires an ImageJ macro and non-standard text file format. The process could be simplified if Cellpose could export the .roi files and zipped .roi collections that ImageJ supports natively, which shouldn't be hard to implement with help from the [roifile](https://github.com/cgohlke/roifile/) library.

carsen-stringer:
thanks for the pointer! we will add this to our to-do list

PR #643, opened 2023-01-24
Export ROIs directly to ImageJ

mrariden:
Allows exporting outlines to .zip archive of .roi files using roifile ImageJ as mention in #581. This introduces a new dependency, [roifile](https://github.com/cgohlke/roifile). 

- Left the "Save Outlines" functionality as it was both in the GUI and in the command line args
- Added `--save_rois` command line flag
- Added GUI button similar to existing "save outlines" functionality

Did not update docs (how/should I do this?)

### DIFF
--- a/cellpose/__main__.py
+++ b/cellpose/__main__.py
@@ -20,7 +20,7 @@ except Exception as err:
     
 import logging
 
-
+# settings re-grouped a bit
 def main():
     """ Run cellpose from command line
     """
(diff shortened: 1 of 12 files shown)


carsen-stringer:
Thanks you so much, this looks great and super useful! I just need to test it, will get to it soon. The documentation can be added here: https://github.com/MouseLand/cellpose/blob/main/docs/outputs.rst

mrariden:
Cellpose segmentation:
image

Save the ROIs:
Going into CP GUI -> save outlines as .zip 

Use the ROIs in FIJI/ImageJ:
1) Open image
2) Import ROIs using File -> open. Recent versions will autodetect that the file is an roi archive
3) Click 'show all' to display the rois on the image. 
image

Note that the `roifile` library appends rois to existing archives. So, if the user removes rois after saving the removed rois will still be there when the user saves again. I circumvented this by deleting the file before writing, but this may not be ideal for images with many ROIs.

I added ctrl+R shortcut for the function in the GUI, and updated relevant docs.

carsen-stringer:
thanks this looks great!

Issue #646, opened 2023-01-27
[BUG] undefined functions for low torch versions


FrickTobias:
**Describe the bug**
If using torch<1.10 CellPose will throw an error when trying to call `torch.cuda.mem_get_info()` as this was introduced in torch 1.10. Since the setup.py script specified `torch<=1.6` this will crash CellPose when run for lower PyTorch versions.

**To Reproduce**
Run the following
```
pip install cellpose torch==1.9
cellpose large_image.tif
```
This will create this error `'torch.cuda' has no attribute 'mem_get_info'`

NB: The image must be large enough to get thrown into the 

**Solution**

Either revoke support for torch<1.10 or define a function that fetches gpu memory in different ways depending on which torch version is being run.

I created a [fork](https://github.com/FrickTobias/cellpose/blob/dev/cellpose/dynamics.py#L582) which has a fix for the issue which is compatible with both cases. If no objection is raised I will open a PR when tested so it works. 

**Run log**
```
2023-01-27 02:19:33,706 [INFO] WRITING LOG OUTPUT TO /root/.cellpose/run.log
2023-01-27 02:19:33,706 [INFO] 
cellpose version: 	2.3.dev7+g03e02bc.d20230127 
platform:       	linux 
python version: 	3.8.8 
torch version:  	1.9.0a0+2ecb2c7
2023-01-27 02:19:35,783 [INFO] ** TORCH CUDA version installed and working. **
2023-01-27 02:19:35,783 [INFO] >>>> using GPU
2023-01-27 02:19:35,788 [INFO] >>>> running cellpose on 6 images using chan_to_seg GRAY and chan (opt) NONE
2023-01-27 02:19:35,788 [INFO] >>>> loading model training-images/cycle-2/models/2023-01-25.livecell.A549.E-and-M.4-fovs-per-well
2023-01-27 02:19:37,666 [INFO] >>>> model diam_mean =  30.000 (ROIs rescaled to this size during training)
2023-01-27 02:19:37,666 [INFO] >>>> model diam_labels =  52.497 (mean diameter of training ROIs)
2023-01-27 02:19:37,666 [INFO] >>>> using diameter 30.000 for all images
2023-01-27 02:19:37,668 [INFO] 0%|          | 0/6 [00:00<?, ?it/s]
^[[D2023-01-27 02:20:47,062 [INFO] 0%|          | 0/6 [01:09<?, ?it/s]
Traceback (most recent call last):
[...]

PR #647, opened 2023-01-27
Add compatability for torch <1.10

FrickTobias:
See issue #646 

TL;DR: torch 1.10 introduced the function `torch.cuda.mem_get_info()` so it crashes for lower torch versions. 

I Added a wrapper function which checks for which version is being run and returns the same information as `torch.cuda.mem_get_info()` but works for both lower and higher than torch 1.10.

Snipped from `dynamics.py`
```
...
if version.parse(torch.__version__) >= version.parse('1.10'):
    # for PyTorch version 1.10 and above
    def mem_info():
        total_mem, used_mem = torch.cuda.mem_get_info()
        return total_mem, used_mem
else:
    # for PyTorch version lower than 1.10
    def mem_info():
        total_mem = torch.cuda.get_device_properties(0).total_memory
        used_mem = torch.cuda.memory_allocated()
        return total_mem, used_mem

if masks.size * 20 > mem_info()[0]:
...
```

I am open to pulling this function definition out of `remove_bad_flow_masks()` (see below), but for now I wrote it inside for minimal invasiveness as I am not read up on cellposes code. 

```
def cuda_mem_info():
    if version.parse(torch.__version__) >= version.parse('1.10'):
        # for PyTorch version 1.10 and above
        total_mem, used_mem = torch.cuda.mem_get_info()
        return total_mem, used_mem
    else:
        # for PyTorch version lower than 1.10
        total_mem = torch.cuda.get_device_properties(0).total_memory
        used_mem = torch.cuda.memory_allocated()
        return total_mem, used_mem
```

### DIFF
--- a/cellpose/dynamics.py
+++ b/cellpose/dynamics.py
@@ -7,7 +7,6 @@ from tqdm import trange
 from numba import njit, float32, int32, vectorize
 import cv2
 import fastremap
-from packaging import version
 
 import logging
 dynamics_logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


carsen-stringer:
thank you so much for implementing this fix! rather than adding a new dependency, could we use string parsing like in setup.py?
https://github.com/MouseLand/cellpose/blob/03e02bca6abf15c4532088cec62a927923a8bf03/setup.py#L37

if you're able to add this change that's great, if not I can modify the PR. thanks again for the help

FrickTobias:
Sure I can fix it.

The reason I parsed it using a module is since builds can include letters and it handles comparion of all of the major/minor/patch tags. 

I'll add a check for major == 1 as well as otherwise future builds of torch 2.0 might get caught in this.

carsen-stringer:
thank you! and good point re: v2.0 we'll need to change that in our setup.py as well

PR #674, opened 2023-03-09
Increase logo image resolution

orena1:
I've upscaled the logo resolution, if you like it.


marius10p:
Thanks, but I think it is appropriate for the logo to be pixelated, given that Cellpose analyzes images of pixels.

PR #680, opened 2023-04-01
Fix #679 - Train sizemodel on dataset with partially empty masks 

### LINKED ISSUE #679: [BUG] IndexError when training sizemodel on dataset with empty masks
**Describe the bug**
While trying to train a `SizeModel` myself, I encountered runtime errors in the `train()` method.

**To Reproduce**
Here is a sample to reproduce this error. For the sake of simplicity, this is just artificial data here.
The first element of `labels` is a full-zero matrix (-> no masks)

```
model = models.Cellpose(model_type='cyto')
image_shape = (3, 256, 256)
images = [np.ones(image_shape), np.ones(image_shape)]
labels = [np.zeros(image_shape), np.ones(image_shape)]
model.sz.train(images, labels, n_epochs=2, batch_size=1)
```

**Run log**
The stacktrace points to `cellpose/models.py`
```
# remove images with no masks
for i in range(len(diam_train)):
    if diam_train[i]==0.0:
>       del train_data[i]
E       IndexError: list assignment index out of range
```

jraufeisen:
I encountered #679 while trying to train a SizeModel on my own dataset.
Please allow me to propose a fix via this PR. A testcase is included in `test_train.py`.

From what I see, there are two small issues
* `del train_data[i]` while iterating over the list can cause the index to go out of bounds.
* `del feat` is run outside the for-loop of iterating over batches. Arguably, this should not be a problem for real world datasets. The sample testcase just happened to catch that as well.

Thanks for maintaining this repository!

### DIFF
--- a/cellpose/models.py
+++ b/cellpose/models.py
@@ -1006,15 +1006,15 @@ class SizeModel():
             diam_test = np.array([utils.diameters(lbl)[0] for lbl in test_labels])
         
         # remove images with no masks
+        indices_to_be_removed = []
         for i in range(len(diam_train)):
-            if diam_train[i]==0.0:
-                del train_data[i]
-                del train_labels[i]
-        if run_test:
-            for i in range(len(diam_test)):
-                if diam_test[i]==0.0:
-                    del test_data[i]
-                    del test_labels[i]
+            if diam_train[i] == 0.0:
+                indices_to_be_removed.append(i)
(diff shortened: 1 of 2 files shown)


carsen-stringer:
thanks @jraufeisen , we are not testing this code so we are not accepting any modifications at this time, we recommend users train the size model with non-empty images

PR #689, opened 2023-04-08
Numpy optimizations

ajinkya-kulkarni:


carsen-stringer:
thanks, what is the advantage of these changes? this is not a primary function

ajinkya-kulkarni:
Using `zeros_like` instead of `zeros` is I think a better implementation as it takes care of making type consistent arrays and may reduce memory usage in certain cases. As I said, it is a minor optimization, but can be useful nevertheless.

carsen-stringer:
thanks, I didn't know that. since this function is used for replicating the paper and doesn't have tests for it, I'd rather be safe and not change it, but thanks for the info

PR #697, opened 2023-04-29
Auto cli docs

mrariden:
Per #698, this allows for the CLI docs to be autogenerated on RTD. 

The argparse function had to be refactored to a new file to avoid mocking imports and get sphinx-argparse to work correctly. There's an open PR to fix that issue in sphinx-argparse: https://github.com/ashb/sphinx-argparse/issues/9#issue-1097057823, but this is a simple workaround.

I also updated some of the docs to link between the pages and deleted the makefiles since they're not used by RTD.


carsen-stringer:
thanks @mrariden this looks really nice, this is a great improvement!

Issue #447, opened 2022-02-21
Increasing Cellpose throughput


VarIr:
Dear Cellpose team, thanks for your great tool.

We love the quality of segmentation we obtain by using `Cellpose`. Currently, we are trying to figure out whether we can increase throughput to saturate our GPUs, without having to run multiple instances of Cellpose.

We use roughly the following code:
``` Python
from cellpose import io, models
model = models.Cellpose(gpu=True, model_type='nuclei')
images = [io.imread(x) for x in paths_of_2D_images]
_ = model.eval(images, ..., batch_size=10000)
```

Apparently, the `batch_size` parameter does not change throughput (and for `10000` I'd actually expect OOM errors). If I understand Cellpose' execution flow correctly [e.g., here](https://github.com/MouseLand/cellpose/blob/8a5360a7b7db7472d5dd21bd2cd5d050e421fea5/cellpose/models.py#L570), a batch is processed so that the `model.eval()`  is recursively called and each image in the batch processed individually (`batch_size` variable is not used in the [recursion anchor](https://github.com/MouseLand/cellpose/blob/8a5360a7b7db7472d5dd21bd2cd5d050e421fea5/cellpose/models.py#L614)).

1. Is it possible to perform segmentation on whole minibatches simultaneously, instead of iterating through individual images? If so, this should result in significantly better performance.

2. Orthogonally, in other parts of the code, I believe performance can be improved by simpler means. For example, the [`UnetModel._run_tiled(()`](https://github.com/MouseLand/cellpose/blob/8a5360a7b7db7472d5dd21bd2cd5d050e421fea5/cellpose/core.py#L578)  iterates over tiles of images that are extracted by [fancy indexing](https://numpy.org/doc/stable/user/basics.indexing.html#advanced-indexing) which creates copies of data, instead of views.
The following should result in slightly faster execution and lower memory footprint (`slice` instead of `np.arange`):
``` python
for k in range(niter):
    irange = slice(batch_size*k, batch_size*k+batch_size)
    y0, style = self.network(IMG[irange], return_conv=return_conv)
    y[irange] = y0
```
(I can create a PR for this, if there is interest).

kevinjohncutler:
@VarIr that's great! Have you quantified the performance difference with the simple change (2) that you proposed?

VarIr:
Not yet, this just caught my eye while reading the code. I'll have a closer look and actually implement & benchmark this.

VarIr:
[...]

PR #709, opened 2023-05-17
replace np.arrange with slice to improve performance

sophiamaedler:
implements speed increase mentioned in point 2 of #447


mrariden (commented):
Looks good to me :+1: 

@carsen-stringer can you review and merge?

Issue #704, opened 2023-05-09
Saving .txt files via GUI or CLI takes extremely long


yassinharim:
I am using Cellpose in an ImageJ macro workflow for large scans of entire developing mouse brains (~20-25k cells counted). In the macro, my images are first processed, then I call Cellpose via the CLI to segment the individual channels using my pretrained models, then I export the masks as a .txt, import them back into ImageJ using the provided python script and then perform colocalisation analysis in ImageJ. 

My issue is that writing the .txt files takes extremely long - in case of an image with ~25k cells, it is around 1h on a machine with i7-11700K, RTX 3060 Ti and NVMe SSD storage. The time is pretty much identical when exporting via the menu item in the GUI. 
Of course we can run the macro overnight, but I'm just wondering why it takes so long to write the file. 

Is there anything that can be done to speed up this export? Or will the direct .roi export submitted in #643 (I'm really looking forward to this functionality!) already be faster?

Update: I manually implemented the changes from #643 and compared exporting times - .txt took 49 minutes, while .roi export completed after 44 minutes. So there doesn't seem to be a large difference between those methods. In the entire workflow, of course, I can save some time because I don't need to run the ROI importer script anymore. But an overall faster export would still be great. 

Thanks a lot!

yassinharim:
I dug a bit more into the code and I found that the function that needs much, much more time with bigger images is `cv2.findContours()`. So I guess this is outside the hands of Cellpose...

mrariden:
@yassinharim I experimented with adding multithreading to the `cv2.findContours()` call and saw up to a 2x speed up. The speedup will be system/hardware dependent. 

@carsen-stringer does this sound like something I should finish and implement?

carsen-stringer:
Thanks @mrariden yes please do! We will still want to check speed up on windows if you haven't already, sometimes multithreading can be a bit weird there

yassinharim:
This sounds great, thank you so much! @mrariden if you would be willing to share the required changes to enable multithreading already, I would be very happy to implement and test on a few different Windows setups and report back.

mrariden:
@yassinharim that would be great! Either clone #712 locally and run `pip install -e .[gui]` in the cloned directory. Or `pip install git+https://github.com/mrariden/cellpose.git`, cloning locally isn't necessary for the latter.

yassinharim:
[...]

PR #712, opened 2023-05-19
add multiprocessing to outlines_list

mrariden:
Per #704, saving outlines takes a long time to find contours for each mask. To speed up the process I added multiprocessing to process the masks in parallel. 

Tested on Ubuntu 22, needs to be tested on Windows and Mac.

### DIFF
--- a/cellpose/utils.py
+++ b/cellpose/utils.py
@@ -9,6 +9,7 @@ from scipy.stats import gmean
 import numpy as np
 import colorsys
 import io
+from multiprocessing import Pool, cpu_count
 
 from . import metrics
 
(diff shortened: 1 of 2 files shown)


mrariden:
@yassinharim reports success on Windows 11 (https://github.com/MouseLand/cellpose/issues/704#issuecomment-1554578176). Will implement tests so GH checks have chance to catch multithreading weirdness.

mrariden:
test passing, good for review

carsen-stringer:
I'm wondering if we should check how big the image is before doing this, if the image is small with few masks then starting the pool is slower

Issue #703, opened 2023-05-08
[FEATURE] Disable autosave in GUI


pvtodorov:
**Is your feature request related to a problem? Please describe.**
I would like to not autosave my masks after every manual edit in the Cellpose GUI. This appears to write a massive GB-sized file each time I edit a single mask. This is slowing me down as Cellpose writes and I am concered it is putting too many write operations on my laptop's SSD.

**Describe the solution you'd like**
A flag to disable autosave when starting up the gui.

carsen-stringer:
@mrariden maybe we can integrate this request with your suggestion to decrease the size of the saved files?

Issue #501, opened 2022-04-14
Deleting multiple labels in Cellpose


Eddymorphling:
Hello,
Thank for you the latest Cellpose release that has allowed me to self-label and train to create models on my own images. This has been immensely help as the my predictions are more accurate.

A quick question, currently I am using the Cellpose GUI and doing "CTRL + left click" to delete labels that are created with the cyto2 model. Is it possible to delete multiple labels at the same time from a larger area by creating a small region and then CTRL+left click on ROI to delete al labels inside the ROI? The cyto2 model creates a lot of false predictions that I am currently deleting one by one when self-training.

Eddymorphling:
Just to add one more thing - when self training in the GUI, would it also be possible to choose the `initial model` to be one of a previously self-trained model? Currently the options are limited to the models found in the `model zoo` in Cellpose.

Thanks again for the awesome tool/interface!

andrewbux:
I know this is comment is a year old, so I was wondering if this feature is currently in the works of being implemented? If so, that is great. If not, I would like to bring it back into discussion as a possible feature. I also have a lot of small inaccurate masks within a large image that are tedious to delete one at a time. This feature would greatly reduce the workload to manually fix each segmentation.

mrariden:
Should be addressed in #718 and the newly added 'disable autosave in the dropdown menus' (#749).

PR #718, opened 2023-05-24
Delete multiple ROIs in GUI

mrariden:
### Add functionality to GUI to allow 'batch' deletion of masks:

Current implementation is a button that starts batch deletion mode and masks can be selected/deselected until pressing the 'done' button. Once the 'done' button is pressed, the masks are removed from the relevant state variables and the mask is saved once. 

### Task list: 
- [x] Add button(s) to start/stop batch deletion mode
- [x] Add function(s) to connect buttons
- [x] Allow region selection of multiple ROIs
- [x] ~~Write tests?~~
- [x] Write documentation with images (maybe video?)
- [ ] User testing on 
    - [x] Linux
    - [x] Windows
    - [ ] Mac
 - [x] Test with 3d image stack 
 - [x] Verify that `manual` tags are properly updated

### Addresses:
- #703 
- #501

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -19,17 +19,21 @@ jobs:
     name: ${{ matrix.platform }} py${{ matrix.python-version }}
     runs-on: ${{ matrix.platform }}
     strategy:
-      fail-fast: false
+      fail-fast: true
       matrix:
         platform: [ubuntu-latest, windows-latest, macos-latest]
         python-version: [3.8, 3.9, "3.10"]
 
     steps:
-      - uses: actions/checkout@v2
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
+      - uses: conda-incubator/setup-miniconda@v2
         with:
-#           auto-activate-base: true
+          miniconda-version: "latest"
+          auto-activate-base: true
+          activate-environment: ""
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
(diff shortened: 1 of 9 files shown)


mrariden:
Buttons now working. When button is clicked, the user can select masks (currently by individually clicking, but will implement region selection) and delete the selected masks with the done button.

Before:
image

Selection:
image

After deletion:
image

mrariden:
All test look good, just need final testing on Mac & Windows.

mrariden:
@LeireIglesiasC This feature is not currently merged into the main branch of cellpose, and that's why you don't see it. For the moment you can speed up deleting cells by disabling autosaving the _seg.npy file (File drop down -> Disable autosave checkbox). You will need to update your cellpose version since this feature was only added last week. You can do that with:

`conda activate cellpose` Activate your environment, assuming you called it cellpose
`pip install cellpose==2.2.3`



_Or_ you can install this version of cellpose (the delete multiple feature) with these commands:

`conda activate cellpose` Activate your environment, assuming you called it cellpose
`pip install git+https://github.com/mrariden/cellpose.git@delete_multiple_roi`

If you do the latter please let us know your experience using the delete multiple feature!

LeireIglesiasC:
Hi @mrariden , Thank you very much for your fast response. I tried both options you gave, the first one works perfectly! I updated the cellpose version and without the automatic saving the whole process is much faster. 
I tried the second option just in case, but since it gave an error after running the command, so I have not used the delete multiple ROIs feature yet sorry.
image

Leire

mrariden:
Ah this merge broke the disable autosave feature by replacing the `save_sets_with_check()` with the original `save_sets()`

PR #749, opened 2023-07-18
Disable gui autosave

mrariden:
For larger files, it would be useful to not save the *_seg.npy file each time a mask is edited. This PR adds a File menu dropdown checkbox that allows the user to turn off autosaving. Additionally, I removed the 'img' data saved to the .npy file to reduce time/space of writing that repeatedly. 

Addresses: 
#703


mrariden:
@carsen-stringer Ready for merge

PR #760, opened 2023-07-28
Qtpy refactor

mrariden:
Replace PyQt5 dependency with more flexible qtpy library. 

Test on:
- OS testing
  - [x] Windows
  - [x] Mac
  - [x] Linux
- [x] implement pip install logic to look if a pyqt install exists (check with rastermap/setup.py)
- [x] Test uninstall pyqt5 followed upgrade pyqt6

### DIFF
--- a/cellpose/gui/gui.py
+++ b/cellpose/gui/gui.py
@@ -7,11 +7,10 @@ import gc
 from natsort import natsorted
 from tqdm import tqdm, trange
 
-import PyQt5
-from PyQt5 import QtGui, QtCore, Qt, QtWidgets
+from qtpy import QtGui, QtCore, QtWidgets
 from superqt import QRangeSlider
 from qtpy.QtCore import Qt as Qtp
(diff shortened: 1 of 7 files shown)


mrariden:
codecov is upset about decreased coverage, but tests are passing. 

GUI test on ubuntu linux works for me

mrariden:
I updated the install to use pyqt6 but allow pyqt5. I was getting an error related to qt not finding a dependency: 

```
qt.core.plugin.loader: QLibraryPrivate::loadPlugin failed on "/<...>/site-packages/PyQt6/Qt6/plugins/platforms/libqxcb.so" : "Cannot load library /<...>/site-packages/PyQt6/Qt6/plugins/platforms/libqxcb.so: (libxcb-cursor.so.0: cannot open shared object file: No such file or directory)"
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: vkkhrdisplay, wayland-egl, wayland, minimal, eglfs, offscreen, minimalegl, linuxfb, xcb, vnc.

Aborted (core dumped)
```

which I fixed by finding the missing package using `ldd` where the `libqxcb.so` file was to identify packages that weren't found. Finally `sudo apt install libxcb-cursor-dev` solved the issue. 

Not sure how big of an issue that will be for other linux users

Issue #772, opened 2023-08-25
[BUG] min_size argument unused


SimonBon:
Hello everyone, 

I use Cellpose to segment images of DAPI stained single cells and, sometimes we have some artefacts in the images which made me want to use the min_size argument of the `.eval` function of `CellposeModel`. After trying out some values and plotting the results I realised that changing the value did not filter out the small segmentation masks. 

What I do is I have a grayscale image that I pass to the eval function using the following line: 

`masks, _, _ = model.eval(img[...,2], diameter=60, min_size=2000)`

If I go into the function that I assume to be responsible to remove small segmentations I see the following after using the following print statement (`print(npix, min_size)`):

...
1983 15
2371 15
3218 15
2663 15
1762 15
3247 15
1649 15
...

I the working solution for me right know is to pass the min_size argument for function `dynamics.compute_masks` on line 652 in module `cellpose.models`

without fix: 
![Pasted Graphic 3](https://github.com/MouseLand/cellpose/assets/62474133/04c5395a-750c-4d01-a1df-0a7717e49fa3)

with fix:
![Pasted Graphic 2](https://github.com/MouseLand/cellpose/assets/62474133/b35a4f40-d014-4a2b-b660-efcb070f96bf)

Thanks for your help!

mrariden:
@SimonBon thanks for identifying and testing a solution. Do you want to make a PR with the changes so you'll get credit in the release notes?

SimonBon:
@mrariden done so, see #773. Thanks

PR #773, opened 2023-08-28
fifxing min_size bug usage in models.py

SimonBon:
Hello as stated in #772 the min_size argument was not used in the dynamics.compute_masks function. Therefore here is the fix.


mrariden (approved):

mrariden:
@carsen-stringer ready for merge

Issue #774, opened 2023-08-29
[BUG] - resample=False causes Value Error


jimmib78:
**Describe the bug**
A Value Error Occurs on some images `ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 2 dimensions. The detected shape was (59, 2) + inhomogeneous part.`
The issue appears if 'resample=False' and there is an image_set that contains no identifiable cells.
If resample is left to the default (True), no error occurs.  I would like to leave this as false as the images are already being downsampled and I having to set diameter=90 to prevent the over-segmentation of larger cells. _Of note I tried using diameter to optimise different sized images (different resolution), but the results are very different to if I downsample the image and use the same  diameter._ 

**To Reproduce**
1. Run cellpose from Jupyter Notebook on blank image
2. set resample=False

**Run log**
```
Frame 58: 1 trajectories present.
processing file: 230823_L1_Akt-GAB-SHP_CS1_Series_10.tiff
2023-08-29 16:36:38,152 [INFO] >>>> loading model Z:/Proj_ERK/Adaptors + Akt/Training 2/models/CP_TIRF10_100x
2023-08-29 16:36:38,153 [INFO] ** TORCH CUDA version installed and working. **
2023-08-29 16:36:38,154 [INFO] >>>> using GPU
2023-08-29 16:36:38,349 [INFO] >>>> model diam_mean =  30.000 (ROIs rescaled to this size during training)
2023-08-29 16:36:38,350 [INFO] >>>> model diam_labels =  56.894 (mean diameter of training ROIs)
2023-08-29 16:36:38,451 [INFO] multi-stack tiff read in as having 59 planes 1 channels
2023-08-29 16:36:38,459 [INFO] 0%|          | 0/59 [00:00<?, ?it/s]
2023-08-29 16:36:38,568 [INFO] 7%|6         | 4/59 [00:00<00:01, 37.03it/s]
2023-08-29 16:36:38,678 [INFO] 14%|#3        | 8/59 [00:00<00:01, 36.63it/s]
2023-08-29 16:36:38,784 [INFO] 27%|##7       | 16/59 [00:00<00:00, 54.07it/s]
2023-08-29 16:36:38,894 [INFO] 42%|####2     | 25/59 [00:00<00:00, 65.16it/s]
2023-08-29 16:36:39,005 [INFO] 58%|#####7    | 34/59 [00:00<00:00, 71.21it/s]
[...]

PR #775, opened 2023-09-01
Scale flows to correct shape

mrariden:
Per #774, when `resize` is not `None`, masks get resized but the flows don't in `dynamics.compute_masks()`. This isn't immediately an issue, except that if no putative cellpix are found ([here](https://github.com/MouseLand/cellpose/blob/95f8d9822e14f7fb5b204c8f302f3f5d71a86d3d/cellpose/dynamics.py#L735C1-L740C31)), zero-filled arrays of the original size are returned. 

The result is that, depending on if putative cellpix are found, iterating through the images ([here](https://github.com/MouseLand/cellpose/blob/95f8d9822e14f7fb5b204c8f302f3f5d71a86d3d/cellpose/models.py#L655C5-L655C5)) produces a list of inconsistent np array shapes. Finally, trying to convert the list-of-arrays into a stack of arrays fails at [line 663](https://github.com/MouseLand/cellpose/blob/95f8d9822e14f7fb5b204c8f302f3f5d71a86d3d/cellpose/models.py#L663C1-L663C32). 

This fix reshapes the flows to the correct shape.

Should we think about tests that address this?

### DIFF
--- a/cellpose/dynamics.py
+++ b/cellpose/dynamics.py
@@ -718,11 +718,29 @@ def get_masks(p, iscell=None, rpad=20):
     M0 = np.reshape(M0, shape0)
     return M0
 
+def resize_and_compute_masks(dP, cellprob, p=None, niter=200,
+                                cellprob_threshold=0.0,
+                                flow_threshold=0.4, interp=True, do_3D=False,
+                                min_size=15, resize=None,
+                                use_gpu=False, device=None):
+    """ compute masks using dynamics from dP, cellprob, and boundary """
+    mask, p = compute_masks(dP, cellprob, p=p, niter=niter,
+                            cellprob_threshold=cellprob_threshold,
+                            flow_threshold=flow_threshold, interp=interp,
+                            do_3D=do_3D, min_size=min_size,
+                            use_gpu=use_gpu, device=device)
+
+    if resize is not None:
+        mask = transforms.resize_image(mask, resize[0], resize[1], interpolation=cv2.INTER_NEAREST)
+        p = transforms.resize_image(p, resize[0], resize[1], interpolation=cv2.INTER_NEAREST)
+
+    return mask, p
+
+
 def compute_masks(dP, cellprob, p=None, niter=200, 
                    cellprob_threshold=0.0,
                    flow_threshold=0.4, interp=True, do_3D=False, 
-                   min_size=15, resize=None, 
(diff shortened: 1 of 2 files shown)


carsen-stringer:
technically the correct thing to do is to leave them in the unresized form, so let's return a resized array if cellpix is none

mrariden:
I think I see, my solution works but it shouldn't because compute_masks should respect the `resize` or `rescale` arguments? (I am still unsure what the convention is for `rescale`, vs `resize`, `reshape`, `resample`.)

Here's the data in case I didn't describe it well enough before. This is the state 5 loops through the iterator on [line 655 in _run_cp()](https://github.com/MouseLand/cellpose/blob/a097ecc931ffb8abb44f99579e4f67edde5ff42b/cellpose/models.py#L655C4-L656C23)

The input dP being processed has a shape of (2, 59, 77, 77) after tiling:
```python
>>>[dPi.shape for dPi in dP[0][0:5]]
[(77, 77), (77, 77), (77, 77), (77, 77), (77, 77)]

>>>[dPi.shape for dPi in dP[1][0:5]]
[(77, 77), (77, 77), (77, 77), (77, 77), (77, 77)]
```

And the output masks are reshaped because of [this line](https://github.com/MouseLand/cellpose/blob/a097ecc931ffb8abb44f99579e4f67edde5ff42b/cellpose/dynamics.py#L761C1-L762C1): 
```python
>>>[m.shape for m in masks]
[(232, 232), (232, 232), (232, 232), (232, 232), (232, 232)]
```
But the `p` array is not reshaped and has inconsistent size as a result:
```python
>>>[pi.shape for pi in p]
[(2, 232, 232), (2, 232, 232), (2, 77, 77), (2, 232, 232), (2, 77, 77)]
```


So, I have two thoughts, (1) should the resize be the opposite [here](https://github.com/MouseLand/cellpose/blob/a097ecc931ffb8abb44f99579e4f67edde5ff42b/cellpose/models.py#L654C2-L654C71) to be `resize = [shape[1], shape[2]] if resample else None` (remove the `not`) ?  And (2), I think we should remove the resizing from compute_masks() since that seems to be a different function.

mrariden:
Okay new solution refactors `dynamics.compute_masks()` to `dynamics.reshape_and_compute_masks()` which calls `dynamics.compute_masks()` which has no internal reshaping behavior.

PR #781, opened 2023-09-13
Simon bon min size fix

carsen-stringer:
edit of #773


carsen-stringer:
@SimonBon (cc @mrariden ) there needs to be an if statement here for 3D stitching -- if you are stitching after then the min_size is not used.

carsen-stringer:
thank you for the pull request @SimonBon !

Issue #735, opened 2023-06-22
Large memory usage when running through CLI


WeilerP:
@carsen-stringer, is there a guideline on the expected memory usage? I was following [this](https://www.10xgenomics.com/resources/analysis-guides/performing-3d-nucleus-segmentation-with-cellpose-and-generating-a-feature-cell-matrix) guide from 10x to process a Xenium sample but I keep running out of memory when running cellpose with the CLI. So far, I requested up to 800GB of RAM but still ran out of memory.

To extract and save the stack level of interest, I am using

```python
import tifffile


def get_tif_image(fpath: str, stack_level: int):
    with tifffile.TiffFile(fpath) as tif:
        image = tif.series[0].levels[stack_level].asarray()

    return image


def write_tif_image(image, fpath: str):
    tifffile.imwrite(
        fpath,
        image,
        photometric='minisblack',
        dtype='uint16',
        tile=(1024, 1024),
        compression='JPEG_2000_LOSSY',
        metadata={'axes': 'ZYX'},
    )
```

which gives me an image of size `(12, 10208, 7814)`. I then run cellpose via the CLI using

```
python -m cellpose --dir ${output_dir} --pretrained_model nuclei --chan 0 --chan2 0 --img_filter _morphology.ome --diameter ${pixel_diameter} --do_3D --save_tif --use_gpu --verbose
```

Is there something inherently wrong with this pipeline that would explain the large memory usage?

When I am running everything through a Python script, I can successfully run the segmentation with a peak memory usage of less than 150GB.

```python
import tifffile

from cellpose import io, models


def get_image(file_path):
    return tifffile.imread(file_path)


if __name__ == "__main__":
    file_path = DATA_DIR / "melanoma" / "region_4" / f"level_{STACK_LEVEL}" / f"level_{STACK_LEVEL}_morphology.ome.tif"
    image = get_image(file_path)

    model = models.Cellpose(model_type="nuclei", gpu=True)

    channels = [0, 0]
    masks, flows, _, diameters = model.eval(image, channels=channels, diameter=10, do_3D=True, progress=True)

    io.masks_flows_to_seg(images=image, masks=masks, flows=flows, diams=diameters, file_names=file_path, channels=channels)
    io.save_masks(images=image, masks=masks, flows=flows, file_names=file_path, png=False, tif=True, channels=channels)

```

I am running everything in Python 3.9 with

```pytb
tifffile==2023.4.12
cellpose==2.2.2
torch==2.0.1
```

[...]

PR #793, opened 2023-09-29
Speedy improvements

mrariden:
When using the `--save_png` flag, the `io.save_masks()` method attempts to make a PNG of the original image/predicted outlines/predicted masks/cell pose. This takes a lot of time, especially for large images. I've added a default flag to disable the block in `io.save_masks()` that makes the matplotlib figure. I've exposed this flag to the CLI and API. 

Alternatively, the image data could be reshaped manually, rather than letting matplotlib do the work. (@carsen-stringer)

Example CLI data after setting default behavior to skip matplotlib: 
- 10 Mb image, 2600x1500
  - Base runtime: __98__ seconds
  - With `--save_png`: __105__ seconds
  - With `--save_png` and `--save_mpl`: __121__ seconds 
- 100 Mb image, 5700x6500
  - Base runtime: 
  - With `--save_png`: __668__ seconds
  - With `--save_png` and `--save_mpl`: __1562__ seconds 

This also is related to the OOM issues #735, since the memory spikes at the end of matplotlib figure creation.

### DIFF
--- a/cellpose/__main__.py
+++ b/cellpose/__main__.py
@@ -182,7 +182,7 @@ def main():
                     io.save_masks(image, masks, flows, image_name, png=args.save_png, tif=args.save_tif,
                                   save_flows=args.save_flows,save_outlines=args.save_outlines,
                                   save_ncolor=args.save_ncolor,dir_above=args.dir_above,savedir=args.savedir,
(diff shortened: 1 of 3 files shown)


mrariden:
Test initially not passing due to persistent GH actions weirdness. (Environment wasn't able to be created). Passing now. @carsen-stringer

PR #807, opened 2023-10-23
Test for Python 3.11 support

harryzcy:
Python 3.11 seems to be working just fine. Python 3.12 doesn't work because `numba` doesn't support it yet (tracked here: numba/numba#9197).


carsen-stringer:
sweet thanks for this!

Issue #789, opened 2023-09-27
ValueError: sero-size array to reduction operation minimum which has no identity [BUG]


LeireIglesiasC:
Hi,

Recently I encountered an error while saving .zip files of the ROIs segmented for an image in cellpose 2.2.3. 

The images (spheroid-like cells) are saved as .TIFF. To segment them, they are individually dragged into the GUI, then you adjust the cell diameter and click on the "cyto" model zoo to segment. The segmentation part works since the masks appear over the cells. There are usually also some masks to correct, so, after editing (deleting and drawing spheres), the final masks are saved as: file > save outlines as .zip archive of ROI files for ImageJ (or Ctr + R). This process works usually very nicely (a bit slow sometimes), but it saves properly the .zip file in the file location of your original image.

However, today I encountered a problem in some images since the .zip file with the ROIs does not save. This error occurs both, when you click save directly after mask obtention with no ROI editing (see code below), and also when saving after ROI editing. I am not very sure when exactly this error occurs, because with some images it saves properly without an error, and then with other images it gives the error. Also, sometimes with the same image it seems to work and if you try again it does not.

This is the error shown: 
`(base) C:\Users\ruberni>conda activate cellpose

(cellpose) C:\Users\ruberni>python -m cellpose
2023-09-27 15:45:38,478 [INFO] WRITING LOG OUTPUT TO C:\Users\ruberni\.cellpose\run.log
2023-09-27 15:45:38,483 [INFO]
cellpose version:       2.2.3
platform:               win32
python version:         3.8.17
torch version:          2.0.1+cpu
2023-09-27 15:45:38,816 [INFO] TORCH CUDA version not installed/working.
GUI_INFO: loading image: L:/Labs/Niels Ruber/Montages (18-06-23)/1. Organised folders - analysis/151/151-DIV7-Pl2-H1/2305151-151-004-NR-P0-DIV7-Pl2-H1_B03_B03_w1.tif
(2764, 2764, 4)
GUI_INFO: auto-adjust enabled, computing saturation levels
2023-09-27 15:46:09,857 [INFO] >>>> using CPU
2023-09-27 15:46:09,857 [INFO] >> cyto << model set to be used
2023-09-27 15:46:09,967 [INFO] >>>> model diam_mean =  30.000 (ROIs rescaled to this size during training)
2023-09-27 15:46:09,983 [INFO] ~~~ FINDING MASKS ~~~
[...]

Issue #708, opened 2023-05-16
Empty line in roi txt file causing bug in imagej_roi_converter.py[BUG]


iagopinal:
**Describe the bug**
The .txt file with the ROIs generates some empty lines that cause an error in line 19 of the imagej_roi_converter.py (xy = map(int, line.rstrip().split(",")))

**To Reproduce**
I modified slightly the imagej_roi_converter.py to be able to input files programmatically. When it tries to input some .txt files with the rois containing an empty line the map function fails and only the rois before that empty line are reported. It has an easy fix which is to ignore those lines:

```
if len(line.strip()) == 0:
        continue
```

If anybody needs to load the rois programmatically, this is the full implementation:
```
#@String my_file

from ij import IJ
from ij.plugin.frame import RoiManager
from ij.gui import PolygonRoi
from ij.gui import Roi
from java.awt import FileDialog

print(my_file)

RM = RoiManager()
rm = RM.getRoiManager()

imp = IJ.getImage()

textfile = open(my_file, "r")
for line in textfile:
    if len(line.strip()) == 0:
        continue
    print(line)
    xy = list(map(int, line.rstrip().split(",")))
    print(xy)
    X = xy[::2]
    Y = xy[1::2]
    imp.setRoi(PolygonRoi(X, Y, Roi.POLYGON))
    # IJ.run(imp, "Convex Hull", "")
    roi = imp.getRoi()
    print(roi)
    rm.addRoi(roi)
textfile.close()
rm.runCommand("Associate", "true")
rm.runCommand("Show All with labels")
```

mrariden:
Looks good @iagopinal, can you make a pull request with the changes?

francodgstn:
I just created a PR with similar solution.
https://github.com/MouseLand/cellpose/pull/828

PR #829, opened 2023-11-19
Skip empty outlines and display a message in the console.

francodgstn:
I needed a quick workaround for the issue #789 (and #708)
It doesn't address the root cause, why some outlines are empty, but at least it allows to export the ImageJ ROI file.

I'm not used to python, you might want to adjust the code.


francodgstn:
@mrariden I see you have the bug #789 assigned, any chance to review my change?

carsen-stringer:
did the other PR fix this?

francodgstn:
Hi, not exactly. The other PR was just to ignore empty line in the imagej_roi_converter.py.
Here the idea is to do not generate those at all and give feedback in the console.

Issue #836, opened 2023-12-07
[FEATURE] Running cellpose on the non-functional channel of suite2p data


landoskape:
**Scientific Goal**
I want to identify cells that express a red fluorescent protein to help labelling which functional ROIs defined by suite2p express the red fluorophore. Since suite2p registers the functional and non-functional channels together, I thought that cellpose would be a good tool for this. 

**Current Solution**
The following steps would work just fine:
1. Run suite2p
2. Use custom script to save ``meanImg_chan2`` (or a different image) as an independent tif or jpeg file.
3. Run cellpose on that image file.

**Possible Solution**
However, it would be nice if cellpose could accept a suite2p ``ops.npy`` file directly, or better yet, accept an entire suite2p folder with multiple planes of ``ops.npy`` files, then save the data to the suite2p folder appropriately. 

**Additional Context**
If this is something that is useful for more people than me and doesn't exist yet I'm happy to share whatever code I write.

carsen-stringer:
thanks landoskape, I think this is more appropriate as an enhancement for suite2p. but we will keep the PR around in case it's useful for others

PR #837, opened 2023-12-07
Allow user to drop suite2p folder into GUI

landoskape:
This is a PR that allows the user to drag and drop an entire suite2p folder to the cellpose GUI and (by default) load the ``meanImg_chan2`` reference images from every plane as a stack. Addresses issue #836.

Feel free to ignore the PR if this seems out of scope for the repo, but I added it in case it might be useful for anyone else.


carsen-stringer:
thanks @landoskape! I've thought about this and I think it's a bit too specialized of a feature but I will leave the PR here for others

PR #839, opened 2023-12-08
Prevent list of trained model names from having repeats

landoskape:
In `add_model()` of the `io` module, the list of user model_names is appended to with the new name, even if it already existed. This is confusing, because in the select custom model section of the GUI, it will repeat the name, even though only one model under that name already exists. This corrects that by checking if the `fname` of the new model exists in `models.get_user_models()` before adding it.


carsen-stringer:
thanks!

PR #840, opened 2023-12-08
Permit pretrained model to be used as baseline

landoskape:
I haven't looked carefully at the training procedure, but it seems like users might want to build off of their previous models without having to put every image they've curated into the same folder.

### DIFF
--- a/cellpose/gui/guiparts.py
+++ b/cellpose/gui/guiparts.py
@@ -10,6 +10,7 @@ from pyqtgraph import functions as fn
 from pyqtgraph import Point
 import numpy as np
 import pathlib, os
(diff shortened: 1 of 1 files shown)


carsen-stringer:
this will cause overfitting, unless the users have trained a model with many many images in the past, so we do not want to enable this from the GUI

mrariden:
@landoskape Yes they are different, review the paper's description of HITL. In short, the ideal is to train the pretrained model only once with your data. Each time you train using HITL you are adding data and creating a new model that is trained starting from the pretrained weights. 

As mentioned, training multiple times will cause overfitting and lose the benefit of the diverse pretrained model.

PR #846, opened 2024-01-18
Update dynamics.py with torch version fix

FerranC96:
`torch.__version__.split(".")` could return more than 3 arguments, which the previous code did not account for.

For example, on my local machine with a fresh `torch `install from Conda, `torch.__version__` returns `'2.0.0.post301'.
With the proposed changes, it will only grab the first 2 items after the "." split, thus eliminating the need for the throwaway variable and also the ValueError that would be raised if the version split returns more than 4 variables.


carsen-stringer:
we are going to remove support for torch < 1.10, thanks though!

FerranC96:
No problem! Looking forward to checking out v3!

PR #862, opened 2024-02-16
added optional params cp_path and logfile_name to the setup_logger fu…

MaxJoas:
As of my understanding, the only way to get information about the loss, is in the ~/.cellpose/run.log file.
For me (and probably for others too) I often start multiple runs in parallel. To have one log file per run, it would be useful to set a custom logging dir and logfile name in the function `logger_setup` in `cellpose/io.py`.
I've added the optional params `cp_path` and `logfile_name` with the default values ".cellpose" and "run.log" (same values that are harcoded in the code)


carsen-stringer:
thanks! I am also adding the train and test loss as outputs from `train.train_seg`

PR #881, opened 2024-03-01
Add .flex to the allowed image file extensions

mmasinas:
FLEX is one of the common file type of microscopy images. It would be great if this can be added to the allowed file type extensions.


carsen-stringer:
thanks!

Issue #762, opened 2023-08-07
[FEATURE] Prediction/training on all available channels


remyeltorro:
Hello,

First of all, I'd like to thank you for the great work. 

I'm highly interested in multimodality and information contained across more than one channel. In that spirit, I wanted to use the --all_channels option to train a custom Cellpose model on all the channels I provide, without restricting myself to only 1-cyto (and 1 optional nucleus channel). I passed the --all_channels tag to the CLI and everything seemed to work fine. 

When I wanted to predict using this new model, on the other hand, I got stuck by an unfulfilled condition at line 504 of the cellpose.models.py file, in the eval function of the CellposeModel. From what I read of the source code, I can do inference with such a model if I pass _model_type=None_ when loading the new model and _channels=None_ when using the eval function. 

I am not allowed to pass channels=None due to this condition, I get "TypeError: object of type 'NoneType' has no len()":

```python
 channels=channels[i] if (len(channels)==len(x) and 
            (isinstance(channels[i], list) or isinstance(channels[i], np.ndarray)) and
             len(channels[i])==2) else channels, 
```
I manually erased the if condition and just passed "channels" and it seems to work. Do you think it would be possible to rework this line in a way that we can actually use Cellpose on any channels without modifying the source code?

Thank you very much for your attention,

mrariden:
@remyeltorro can you share a small complete example that illustrates the problem you're having? I'll be better able to help with the actual code at hand

remyeltorro:
@mrariden  thank you very much, I'll try to elaborate:

The idea: pass two or more 'cyto-like' channels to Cellpose, without performing RGB merging or other preprocessing. The info on how to segment the cells is on more than one channel and I want this as the input to the underlying UNET. I want to keep everything else the same. 

Say I produced a model with the following CLI:

python -m cellpose --use_gpu --train --dir /home/limozin/Desktop/segment_RICM/annotations/dataset/train/augmented/ --test_dir /home/limozin/Desktop/segment_RICM/annotations/dataset/train/test/ --learning_rate 0.001 --weight_decay 0.0001 --n_epochs 5000 --mask_filter _labelled --verbose --no_norm --batch_size 8 **--all_channels**

The model detects the number of channels in my images and trains smoothly, I think it does what I intended, after tracking down in the source code what the --all_channels tag does.

Then if I want to load the model for prediction, it'll look something like this:

[...]

PR #949, opened 2024-05-24
allows --all channels to work in CLI

lacan:
Referencing #762 would this quick fix actually solve the issue? I have not tested it with DenoiseModels. 

Let me know how I can further help if at all

Oli


lacan:
Any update on this?

carsen-stringer:
good catch thanks!

PR #967, opened 2024-07-03
Downgraded `numpy` to be less then version 2.0.0

ajinkya-kulkarni:
Installing `cellpose` sometimes gives errors due to `numpy` version


carsen-stringer:
thanks we've done this already

PR #970, opened 2024-07-10
Fix link in doc

maweigert:
Fixes a html link in the documentation


carsen-stringer:
thanks! I think we now have full M1-3 support out-of-the-box, but let me know if you run into any issues, I will update the instructions

PR #979, opened 2024-07-24
`imagecodecs` no longer compiles with `python 3.8`

ajinkya-kulkarni:
Changed minimum python requirement to 3.9


carsen-stringer:
thanks!

PR #981, opened 2024-07-25
Fix intensity normalizations

qin-yu:
Hi @carsen-stringer!

During the process of packaging the custom Cellpose model for bioimage.io, I identified two potential issues with the intensity normalization implementation in Cellpose. This PR proposes fixes for:

1. The current normalization with the `lowhigh` argument is not channel-wise. In this PR, `lowhigh` now either takes a 2-tuple `(low, high)` for all channels, or a `nchan`-tuple of that for each channel.
2. The current invert mechanism does not necessarily invert the channel to be segmented (`--chan`); instead, it inverts the last channel. Depending on the `--chan2` value, this might not be the intended channel. In this PR, all channels are inverted if `invert`.

<details>

<summary>Original PR message on 25 Jul 2024</summary>

## Original PR

Hi @carsen-stringer!

During the process of packaging the custom Cellpose model for bioimage.io, I identified two potential issues with the intensity normalization implementation in Cellpose. This PR proposes fixes for:

1. The current min-max normalization with the `lowhigh` argument is not channel-wise, and the argument itself might be misused.
2. The current invert mechanism does not necessarily invert the channel to be segmented (`--chan`); instead, it inverts the last channel. Depending on the `--chan2` value, this might not be the intended channel.

Please review these changes as I might have misunderstood some parts.

----

### Fix for Min-Max Normalization by 31bc5befb3573f9dd0cc46db392788fe32c96bcd

For `cellpose.transforms.normalize_img()` (channel-wise normalization), the `lowhigh` option is provided. I assume this corresponds to min-max normalization where `lowhigh` is a tuple representing (new minimum, new maximum).
https://github.com/MouseLand/cellpose/blob/84344b0ceb6eb91e759301fbdad5c9b77a1c881b/cellpose/transforms.py#L600-L601

If `lowhigh` represents (old minimum, old maximum), then normalization is not channel-wise if `lowhigh` has a length of 2:
https://github.com/MouseLand/cellpose/blob/84344b0ceb6eb91e759301fbdad5c9b77a1c881b/cellpose/transforms.py#L624-L626

In either case, the implementation needs fixing:
https://github.com/MouseLand/cellpose/blob/84344b0ceb6eb91e759301fbdad5c9b77a1c881b/cellpose/transforms.py#L640-L643
The current implementation is essentially: `img_norm = (img_norm - lowhigh[0]) / (lowhigh[1] - lowhigh[0])`

This has been fixed by commit 31bc5befb3573f9dd0cc46db392788fe32c96bcd.

----

### Fix for Invert by 91469d1273ee0b3e3e4f614d0e8a80b0587ceb40

For `cellpose.transforms.normalize_img()` (channel-wise normalization), there is an `invert` option for cases where "cells are dark instead of bright":
https://github.com/MouseLand/cellpose/blob/84344b0ceb6eb91e759301fbdad5c9b77a1c881b/cellpose/transforms.py#L598-L599

In Cellpose, the invert function is performed after `reshape()`, which moves the main channel to channel 0 and the auxiliary channel to channel 1. Therefore, the primary use case is to invert the first channel (channel 0) instead of the last one (current implementation):
https://github.com/MouseLand/cellpose/blob/84344b0ceb6eb91e759301fbdad5c9b77a1c881b/cellpose/transforms.py#L666-L667

The variable `c` retains the value from a previous loop, being 1 if `nchan` is 2 or 2 if `nchan` is 3.

This has been fixed by commit 91469d1273ee0b3e3e4f614d0e8a80b0587ceb40, but further improvements can be made to allow for a user-specified list of channels to be inverted.

</details>

### DIFF
--- a/cellpose/transforms.py
+++ b/cellpose/transforms.py
@@ -393,7 +393,7 @@ def move_min_dim(img, force=False):
 
     Args:
         img (ndarray): The input image.
-        force (bool, optional): If True, the minimum dimension will always be moved. 
+        force (bool, optional): If True, the minimum dimension will always be moved.
             Defaults to False.
 
     Returns:
@@ -668,7 +668,7 @@ def normalize_img(img, normalize=True, norm3D=False, invert=False, lowhigh=None,
                                                            c], lower=percentile[0],
                                                   upper=percentile[1], copy=False)
         if (tile_norm_blocksize > 0 or normalize) and invert:
-            img_norm[..., c] = -1 * img_norm[..., c] + 1
+            img_norm[..., 0] = -1 * img_norm[..., 0] + 1
         elif invert:
             error_message = "cannot invert image without normalizing"
             transforms_logger.critical(error_message)
@@ -690,7 +690,7 @@ def resize_image(img0, Ly=None, Lx=None, rsz=None, interpolation=cv2.INTER_LINEA
         Lx (int, optional): Desired width of the resized image. Defaults to None.
         rsz (float, optional): Resize coefficient(s) for the image. If Ly is None, rsz is used. Defaults to None.
         interpolation (int, optional): OpenCV interpolation method. Defaults to cv2.INTER_LINEAR.
-        no_channels (bool, optional): Flag indicating whether to treat the third dimension as a channel. 
+        no_channels (bool, optional): Flag indicating whether to treat the third dimension as a channel.
             Defaults to False.
 
     Returns:
@@ -742,8 +742,8 @@ def pad_image_ND(img0, div=16, extra=1, min_size=None):
         extra (int, optional): Extra padding. Defaults to 1.
         min_size (tuple, optional): Minimum size of the image. Defaults to None.
 
-    Returns:   
-        tuple containing 
+    Returns:
+        tuple containing
             - I (ndarray): Padded image.
             - ysub (ndarray): Y range of pixels in the padded image corresponding to img0.
             - xsub (ndarray): X range of pixels in the padded image corresponding to img0.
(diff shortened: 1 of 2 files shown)


carsen-stringer:
thanks @qin-yu ! the lowhigh is meant to represent the values to set to 0 and 1 (e.g. pixel value 100->0 and 1100->1 corresponds to lowhigh = [100,1100]. I think the fix here would be to check if lowhigh is a list of lists (per channel) or a single list (for all channels). It's a good point that this use-case isn't covered currently.

for the inversion, thanks for catching that, I think it would make sense to apply for all channels?

qin-yu:
> thanks @qin-yu ! the lowhigh is meant to represent the values to set to 0 and 1 (e.g. pixel value 100->0 and 1100->1 corresponds to lowhigh = [100,1100]. I think the fix here would be to check if lowhigh is a list of lists (per channel) or a single list (for all channels). It's a good point that this use-case isn't covered currently.

Thanks for explaining! That makes sense, working on it.

> for the inversion, thanks for catching that, I think it would make sense to apply for all channels?

Yes you are right, inverting all channels by default if `invert == True` seems natural to me. But maybe this could also be either a `bool` or `tuple(bool, ...)`?

carsen-stringer:
Thanks so much, regarding the invert function I think it's fine with all channels, I'm not sure how much it's being used with multichannel images

qin-yu:
> Thanks so much, regarding the invert function I think it's fine with all channels, I'm not sure how much it's being used with multichannel images

You are right. Let's keep it simple

I have finalized the PR, including two fixes along with the corresponding tests. Please take a look when you have a moment! @carsen-stringer

qin-yu:
The failed check seems to be incidental and maybe unrelated to the PR.

PR #995, opened 2024-08-14
Add missing train/test files to _process_train_test() call

kemaleren:
Modify `train_seg()` so that the `train_labels_files` and `test_labels_files` arguments are correctly passed to `_process_train_test()`.


carsen-stringer:
thank you!

PR #1019, opened 2024-09-24
FEATURE: ctrl-c/v to copy the selected mask into the current layer

outofculture:
This feature came up as a useful addition while annotating 3D data. The particulars we picked (which keys are bound, the search strategy) are totally arbitrary, so they're easy to change.


carsen-stringer:
this is a good idea, I need a bit more time to test it, but thank you for the contribution!

outofculture:
this has a bug! I don't really know what `outpix` is for, and I ignored it, but now sometimes when merging two masks, one mask will steal the outline of a completely unrelated mask, and I think that has something to do with this copy feature, but I haven't figured it out yet.

PR #1036, opened 2024-10-30
New 3D mode

mwshinn:
## Summary

We (a subset of cortexlab) were having a lot of trouble with 3D segmentation.  Cells were being either oversplit or missed completely to the point of being unusable.  I came up with the following new 3D segmentation algorithm, which works very well on our noisy data and largely fixes our segmentation issues.

This PR improves the ability to train on 2D planes and segment in 3D.  Cellpose currently has two modes for doing this:

- Stitching mode: Segment all planes independently, and then combine them using an intersection-over-union rule
- 3D mode: Run the same network (or now, with the latest commits, perhaps a different network) on the xy, yz, and xz planes independently, and then sum the results

The PR replaces 3D mode.  While it has similarities to stitching mode, it is most similar to 3D mode, so I will use 3D mode for baseline comparison from here onwards (and because issues with the stitching method are obvious and well-characterised).

So far, this new algorithm has been tested on and substantially improves segmentation on several datasets, and we're now using it as our default segmentation algorithm.  This PR implements the new algorithm in cellpose4-alpha (what I'm calling the latest github version in the "main" branch), even though we have been running it on cellpose 3.0.7.  Since the arguments for "run_3D" changed last week, I rewrote the algorithm for the new function arguments for the sake of this PR, but most testing has been on the version compatible with 3.0.7 (included below).

Currently, this PR works really well on larger filled cells in thick confocal sections (e.g., GFP-filled cells), giving almost 0 erroneous merges and 0 oversplits; and on dense dapi from thick confocal sections, giving roughly 5% erroneous merges and a handful oversplits (which are easy to detect, since they are oversplit into very small cells which can be filtered out).  I would have preferred to do a bit more testing and hyperparameter tuning on this before submitting a PR, but I see you've been working on the 3D mode lately so figured now is better than later.

## Background

The issues we had with 3D mode are:

- The gradient map often skips planes in noisy data.  This results in striation patterns in the 3D gradient map.  A similar issue occurs with the cellprob map.
- When images are noisy, a cell may have high probability in planes N and N+2, but low probability in plane N+1.
- For cells that are not radially symmetric (e.g. pyramidal neurons), training in one orientation will not help when evaluating in a different orientation
- The PSF is very different in xy than in xz and yz.  (This is especially a problem with multichannel images, as the wide psf in z means that the nuclear/dapi stain is visible in most xy planes, even if those planes don't directly intersect with the nucleus, but only in the xz or yz planes which intersect with the nucleus.)
- In many of our low-SNR images, it is not possible to tell by eye from the xz or yz planes whether a cell is present at all.  They are only faintly visible in the xy plane.
- This is a small issue, but because the three dimensions are summed instead of averaged, the cellprob and flow thresholds need a different scale in 3D mode compared to 2D mode.

When datasets are beautiful and all cells have a high SNR, most of these are not much of an issue.  But when the SNR is low and some cells are difficult to discern by eye, these become problematic.

## Summary of changes

This PR makes the following changes:

- It uses a maximum filter on the gradients and cellprobs in the z dimension to avoid skipped planes (avoiding the hash-like pattern in the cellprob and flows for some datasets, which leads to oversplitting in all dimensions)
- It estimates a z gradient using the filtered x and y gradients, eliminating the need to run networks in xz and yz planes
- By using the x and y gradients to estimate the z gradient, it avoids the case where the gradient is positive in x and y but near zero in z (resulting in oversplit cells along the z axis)
- It assumes the xy plane predictions are much more accurate than the xz and yz plane predictions (which tends to be the case, even for humans doing ground-truth segmentation)
- It allows the cellprob/flow thresholds to have the same scale in 2D and 3D mode

## Comparison with cellpose4-alpha changes in 3D mode

Cellpose4-alpha includes (in the code but not yet in the GUI?) a way to run a separate network on the xz and yz ("ortho") planes.  This will help, but will not completely solve the problems above because (a) there may still be striations, which confuse the flow-based segmentation algorithm (b) the xz and yz networks will usually be much worse than the xy network, but in 4-alpha they are weighted equally.  An alternative may be to replace the z-diff operation in the proposed implementation here with a run of the network.  I have implemented this in the PR but not tested it.

The latest version also includes the option for gaussian filtering the flows.  In my tests I found this to be only slightly helpful - by and large, the flow-based segmentation is still getting confused.  The gaussian filters included here (after the maximum filter) are intended only to avoid the plateaus generated by the maximum filter, ensuring that the derivative of the gradient map is not zero.  Perhaps this addresses a different problem: I didn't test the new gaussian filter extensively, but I tended to get fewer pizza-like patterns when I did.

## Room for improvement

- This algorithm can still merge cells that are very close together.  One possibility for enhancement could be to modify the computed z gradient based on the pairwise similarity of the xy gradients across z planes, instead of based on the magnitude.  This may be able to avoid some erroneous merges of cells that are very close not exactly on top of each other.
- I'm not sure the magnitude of the xy gradient is the best signal to use here when determining z.  I chose it because it seemed to capture the contour of the cell well in all dimensions, not for any principled reason.

## Version of this PR for cellpose 3

For visitors seeing this PR but not wanting to use cellpose 4-alpha, add the following function to the end of cellpose/core.py (tested in cellpose 3.0.7):

```python
import scipy.ndimage
def run_3D(net, imgs, batch_size=8, rsz=1.0, anisotropy=None, augment=False, tile=True,
           tile_overlap=0.1, bsize=224, progress=None):
    """
    Run network on image z-stack.

    (faster if augment is False)

    Args:
        imgs (np.ndarray): The input image stack of size [Lz x Ly x Lx x nchan].
        batch_size (int, optional): Number of tiles to run in a batch. Defaults to 8.
        rsz (float, optional): Resize coefficient(s) for image. Defaults to 1.0.
        anisotropy (float, optional): No effect
        augment (bool, optional): Tiles image with overlapping tiles and flips overlapped regions to augment. Defaults to False.
        tile (bool, optional): Tiles image to ensure GPU/CPU memory usage limited (recommended); cannot be turned off for 3D segmentation. Defaults to True.
        tile_overlap (float, optional): Fraction of overlap of tiles when computing flows. Defaults to 0.1.
        bsize (int, optional): Size of tiles to use in pixels [bsize x bsize]. Defaults to 224.
        progress (QProgressBar, optional): pyqt progress bar. Defaults to None.

    Returns:
        y (np.ndarray): output of network, if tiled it is averaged in tile overlaps.Size of [Ly x Lx x 3] or [Lz x Ly x Lx x 3].
            y[...,0] is Y flow; y[...,1] is X flow; y[...,2] is cell probability.
        style (np.ndarray): 1D array of size 256 summarizing the style of the image,if tiled it is averaged over tiles.
    """
    # Run the network once for each z plane
    print("NOTE: running a modified 3D mode")
    xsl = transforms.resize_image(imgs, rsz=rsz)
    shape = imgs.shape
    y, style = cellpose.core.run_net(net, xsl, batch_size=batch_size, augment=augment, tile=tile, bsize=bsize,
                                     tile_overlap=tile_overlap)
    y = transforms.resize_image(y, shape[1], shape[2])
    yflow, xflow, cellprob = y.transpose([3,0,1,2]) # outputs, z dims, y dims, x dims
    del y
    if progress is not None:
        progress.setValue(50)
    # Compute the matitude of the derivative in the xy plane at each point.
    # This is a rough indication of the physical footprint of cell.  However,
    # since network output will be a bit different for each plane, we need to
    # filter to make sure it is smooth across z.  We will use the gradient of
    # this as our pseudo-flow for the z axis.
    dmag = np.sqrt(yflow**2+xflow**2)
    _dmag_grad = np.diff(scipy.ndimage.gaussian_filter(scipy.ndimage.maximum_filter(dmag, [3,1,1]), 3), axis=0)
    del dmag
    dmag_grad = np.concatenate([_dmag_grad, _dmag_grad[[-1]]], axis=0)
    del _dmag_grad
    # Filter with an "absolute maximum" filter followed by a gentle gaussian.
    # The "absolute maximum" filter is like a "maximum" filter but uses the
    # minimum instead of the maximum if it is greater in absolute value.  The
    # gentle gaussian at the end is to avoid plateus that come from max/min
    # operations.  (This could be done with generic_filter but that is much
    # slower than running both maximum_filter and minimum_filter.)
    flows = np.asarray([yflow, xflow])
    _flows_pixel_filt_pos = scipy.ndimage.maximum_filter(flows, [1, 3, 1, 1])
    _flows_pixel_filt_neg = scipy.ndimage.minimum_filter(flows, [1, 3, 1, 1])
    del flows
    flows_pixel_filt = _flows_pixel_filt_neg
    inds = np.abs(_flows_pixel_filt_pos)>np.abs(_flows_pixel_filt_neg)
    flows_pixel_filt[inds] = _flows_pixel_filt_pos[inds]
    del _flows_pixel_filt_pos
    yflow_filt = scipy.ndimage.gaussian_filter(flows_pixel_filt[0], 1)
    xflow_filt = scipy.ndimage.gaussian_filter(flows_pixel_filt[1], 1)
    del flows_pixel_filt
    # Last we need to compute cell probability.  We will just use the network
    # output nearly verbatim, but filtered along the z axis for smoothness (like
    # above).
    cellprob_filt = scipy.ndimage.gaussian_filter(scipy.ndimage.maximum_filter(cellprob, [3,1,1]), [1,1,1])
    # Renormalise the pseudo-flow in z to match the flows in x and y
    norm = np.maximum(np.max(np.abs(yflow)), np.max(np.abs(xflow)))/np.max(np.abs(dmag_grad))
    new_flows_pixel = [dmag_grad*norm, yflow_filt, xflow_filt, cellprob_filt]
    dz = new_flows_pixel[0]/2
    dy = new_flows_pixel[1]/2
    dx = new_flows_pixel[2]/2
    cp = new_flows_pixel[3]/3
    del new_flows_pixel
    ret = np.asarray([[dy, dx, cp], [dz, dx, cp], [dz, dy, cp]]), style
    return ret
```

### DIFF
--- a/cellpose/core.py
+++ b/cellpose/core.py
@@ -10,6 +10,7 @@ from urllib.parse import urlparse
 import tempfile
 import cv2
 from scipy.stats import mode
+import scipy.ndimage
 import fastremap
 from . import transforms, dynamics, utils, plot, metrics, resnet_torch
 
@@ -291,8 +292,6 @@ def run_3D(net, imgs, batch_size=8, augment=False,
     Args:
         imgs (np.ndarray): The input image stack of size [Lz x Ly x Lx x nchan].
         batch_size (int, optional): Number of tiles to run in a batch. Defaults to 8.
-        rsz (float, optional): Resize coefficient(s) for image. Defaults to 1.0.
-        anisotropy (float, optional): for 3D segmentation, optional rescaling factor (e.g. set to 2.0 if Z is sampled half as dense as X or Y). Defaults to None.
         augment (bool, optional): Tiles image with overlapping tiles and flips overlapped regions to augment. Defaults to False.
         tile_overlap (float, optional): Fraction of overlap of tiles when computing flows. Defaults to 0.1.
         bsize (int, optional): Size of tiles to use in pixels [bsize x bsize]. Defaults to 224.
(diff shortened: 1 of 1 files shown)


carsen-stringer:
Hi Max, thanks for the input. This algorithm doesn't use ZX and ZY planes - and thus does not use 3D information - so arguably it's more similar to a stitching algorithm. I ran your code on the 3D volume with ground-truth from the cellpose1 paper and received substantially worse results. I do like the idea of reweighting ZX and ZY for X and Y flows, that could be a parameter we add if you have data where you've tested it and it works well (I've tried it and on the data I've tried it didn't help). I am concerned about using max filtering especially with a circular variable.

I think it would make sense for you to keep a fork of cellpose for the cortexlab with this code for your data with these sorts of issues. The issues you're describing with the data -- variability in expression across planes -- seems like something that may be easier to address as a pre-processing step. We also have new anisotropic deblurring models as well in Cellpose3 which we will release code for soon that may help with your ZY and ZX planes.

image

mwshinn:
Thank you Carsen for taking a look at this and testing it on your data!

The segmentation for the image you showed indeed looks very bad.  I'm assuming this is an xz or yz slice?  The pattern shown in your picture (with the sharp horizontal lines) is similar to what I saw in testing when I returned the wrong matrix transpose from run_3D.  It is odd that it is popping up here and I wonder if it is a bug from when I adapted from 3.0.7 -> 4-alpha.  Is the 3D annotated data from the Cellpose 1 paper available online?  I would like to try it but can't seem to find it on http://www.cellpose.org/dataset.

On data with crystal clear xz and yz planes that look like the xy plane, I would guess this PR would not help or would make segmentation slightly worse.  I attached some images of example data, using a two-channel custom-trained network for gfp+dapi.  As you can see, even though there is 3D structure, the z dimension has very different statistics than the x and y dimensions, so I wouldn't expect a network trained on xy to work on xz or yz. It is actually very impressive that cellpose works at all, even in 2D, on these images with such weak gfp!

XY slice:
xy slice

XZ slice:
xz slice


Here are segmentation renders of the native cellpose 3D and this PR's 3D mode.

Cellpose 3D original:
cellpose 3D originall

This PR 3D mode:
This PR 3D mode

It's not perfect (e.g. the purple cells at the bottom are incorrectly merged) but it is still pretty good.

To show one reason why the native cellpose 3D doesn't work as well here, see the below videos going through the z stack of the x gradient when using native cellpose 3D mode.  You can see stripes throughout, especially in the upper right corner.  This is actually one of the more mild cases as far as stripes are concerned, we usually get more of them going in both the x and y directions, even on higher quality images.


https://github.com/user-attachments/assets/40c6e853-5ca3-4f5e-9ef5-c73e71b07aa5


https://github.com/user-attachments/assets/307fc45a-2f4e-470d-8ab8-69f97e825b89


Also, what did you mean about max filtering a circular variable?  Just to clarify, this method max filters the x and y gradients separately, and uses an "absmax" so you don't get a positive or negative bias.

If this 3D mode ends up not being suitable for cellpose mainline, I'll just maintain the code locally as a monkey patch.  I should also mention that I don't expect this to be merged directly as-is - I assume you don't want to replace the existing run_3D function, so its final home might be something like a "stitch_threshold == -1" option.  Happy to write the code to integrate it in the most appropriate way.

carsen-stringer:
Thanks, hmm in that data it looks like it's difficult to use the green channel. You have retrained a model for that data specifically?

The x and y gradients together create a circular variable, so if you are increasing one via a max filter then you are changing the angle and basically emphasizing the cardinal directions? 

Here's a smaller version of the full tiff that's easy to share: https://www.cellpose.org/static/data/rgb_3D.tif

Thanks, I think it makes sense to keep this as a separate fork for now, since this data seems quite unique.

mwshinn:
For those who stumble across this thread in the future, we ended up simplifying the algorithm described in the PR quite a bit and limiting it to a specific version of Cellpose (to avoid the errors Carsen pointed out above).  We now use this patch: https://github.com/mwshinn/cellpose3dplus

PR #1045, opened 2024-11-11
[Doc] Describe how to manage multithreading

royjacobson:
Spent a bit of time on figuring this out (after accidentally running on all 256 cores of our shared login server 😅) so figured it is worth a bit of documenting.
Please LMK if there's a more appropriate place for this.


ashishUthama:
Might also be worth pointing to 
    https://pytorch.org/docs/stable/generated/torch.set_num_threads.html
(Since that ought to keep up with whatever library PyTorch uses under the hood)

carsen-stringer:
thank you this is great!

Issue #1054, opened 2024-11-18
[BUG] Issue when running with post-release version of PyTorch


MeyerBender:
**Describe the bug**
Running cellpose with PyTorch version `2.5.1.post303` leads to the following error:
```
File "/g/huber/users/meyerben/notebooks/spatialproteomics_pipelines/celestimo/.snakemake/conda/4057a625210cbf20c14eec4e61ec3417_/lib/python3.9/site-packages/spatialproteomics/tl/tool.py", line 86, in cellpose
    masks_pred, _, _, _ = model.eval(
  File "/g/huber/users/meyerben/notebooks/spatialproteomics_pipelines/celestimo/.snakemake/conda/4057a625210cbf20c14eec4e61ec3417_/lib/python3.9/site-packages/cellpose/models.py", line 187, in eval
    diams, _ = self.sz.eval(x, channels=channels, channel_axis=channel_axis,
  File "/g/huber/users/meyerben/notebooks/spatialproteomics_pipelines/celestimo/.snakemake/conda/4057a625210cbf20c14eec4e61ec3417_/lib/python3.9/site-packages/cellpose/models.py", line 776, in eval
    masks = self.cp.eval(
  File "/g/huber/users/meyerben/notebooks/spatialproteomics_pipelines/celestimo/.snakemake/conda/4057a625210cbf20c14eec4e61ec3417_/lib/python3.9/site-packages/cellpose/models.py", line 534, in eval
    masks = self._compute_masks(x.shape, dP, cellprob, flow_threshold=flow_threshold,
  File "/g/huber/users/meyerben/notebooks/spatialproteomics_pipelines/celestimo/.snakemake/conda/4057a625210cbf20c14eec4e61ec3417_/lib/python3.9/site-packages/cellpose/models.py", line 620, in _compute_masks
    outputs = dynamics.resize_and_compute_masks(
  File "/g/huber/users/meyerben/notebooks/spatialproteomics_pipelines/celestimo/.snakemake/conda/4057a625210cbf20c14eec4e61ec3417_/lib/python3.9/site-packages/cellpose/dynamics.py", line 839, in resize_and_compute_masks
    mask = compute_masks(dP, cellprob, niter=niter,
  File "/g/huber/users/meyerben/notebooks/spatialproteomics_pipelines/celestimo/.snakemake/conda/4057a625210cbf20c14eec4e61ec3417_/lib/python3.9/site-packages/cellpose/dynamics.py", line 913, in compute_masks
[...]

PR #1055, opened 2024-11-18
Made PyTorch version check compatible with post-release versions

MeyerBender:
Replaced 
```
major_version, minor_version, _ = torch.__version__.split(".")
```
with 
```
torch_version = torch.__version__.split(".")
major_version, minor_version = torch_version[0], torch_version[1]
```
in `dynamics.py`.
Addresses issue #1054, where cellpose would fail when a post-release version of PyTorch was installed.


carsen-stringer:
sorry I didn't see this, I made the change myself, thanks!

Issue #549, opened 2022-08-10
Cellpose on sparse populations


FrickTobias:
# Issue

I'm trying to use CellPose for segmenting sparse cell populations and I am getting really poor performance. The model only identifies dirt rather than cells. 

Could anybody offer any advice? Or does CellPose not handle these kinds of situations? In either case, does anyone have an idea of why it would not work?

![Screen Shot 2022-08-10 at 14 17 55](https://user-images.githubusercontent.com/17293950/183821500-1393b33a-4d39-419a-b281-582011246b19.png)

## CellPose command: 
```

cellpose --dir . --use_gpu --gpu_device 1 --check_mkl --verbose
```

## More info

I've run CellPose successfully on the same kind of sample with much higher degree of success when having higher cell densities (the image had the same sizes, I'm just showing a cropped version I had laying around).

<img width="231" alt="Screen Shot 2022-08-10 at 14 28 39" src="https://user-images.githubusercontent.com/17293950/183822688-91ae74dc-5d14-4564-b979-e2c3938404f3.png">

I've tried running CellPose 2.0.5, 2.0, 1.0.2.

carsen-stringer:
hmm perhaps the diameter is estimated to be too large? could you try specifying the diameter on the command line? e.g. `--diameter 20`. another option is that we now have models for these types of cells, you can try a model trained on [livecell](https://www.nature.com/articles/s41592-021-01249-6) with `--pretrained_model livecell`

alternatively you could try high pass filtering the image, or some other preprocessing, to remove the lower frequency objects you don't care about

going to close but comment if still not working

FrickTobias:
Since the other data worked better with the same `--diameter 30 ` I'm not sure it should be the issue. 

However, I did try using `livecell` instead which worked much better. Thank you!

PR #1058, opened 2024-11-21
Add norm percentiles as CLI argument

TheBestDestroyer:
Added the arg of norm percentiles to CLI options. Currently it only supports norm percentiles but could be extended to also support the other normalization options.

Base motivation was to enable this feature for solving issues on pictures with applied threshold and sparse data, as it is here a good way to tinker with the values. (Also possible in GUI, but not in CLI).

Loosely related  #549 an the blog article [here](https://forum.image.sc/t/cellpose-missing-most-cells-in-sparsely-populated-image/76444).

### DIFF
--- a/cellpose/__main__.py
+++ b/cellpose/__main__.py
@@ -108,6 +108,13 @@ def main():
             default_model = "cyto3"
             backbone = "default"
 
+        if args.norm_percentile is not None:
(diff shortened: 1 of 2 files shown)


carsen-stringer:
thank you this is helpful!

PR #1077, opened 2024-12-12
train.py: remove superfluous percent sign in f-string

deroulers:
This PR removes the superfluous percent sign in an f-string in `train.py` which caused a `ValueError: Invalid format specifier` when calling `train.train_seg` with `save_each=True`. Thanks!


carsen-stringer:
thanks!

PR #1103, opened 2025-02-10
Fix saving of flow fields and cell probability in GUI

qin-yu:
Hi @carsen-stringer, I noticed that _save_flows() was obsolete while trying to save flows in the GUI. I updated the function and tested it on a 3D TIFF and a 2D PNG. I hope this covers all cases.


qin-yu:
( just realised there’s no notification when saving starts or finishes )

carsen-stringer:
thanks so much! made a few small changes and merged it

PR #1104, opened 2025-02-10
feat: custom output name

antoinemeyer5:
### Description
We have got some images in an input A folder and want to use the *MouseLand/cellpose* models on them. These images have a name that makes them unique and identifiable. After processing the *MouseLand/cellpose* models, we would like to have the results for each image in a folder B with the same file names as the input files. So image named X in folder B corresponds to image X in folder A after calculation by the AI.

*MouseLand/cellpose* team already made that possible thanks to the parameters `output_name`. Unfortunately, it's impossible to set this parameter to an empty string, and that's perfectly normal because we do not want to ever over-write someone's original tif.

### Changes made
I have added the check that a different input (`dir`) and output (`savedir`) folders are properly passed as parameters and if so, then you don't need to add anything to the output file (`output_name`) name.

### Benefits
In cases where the input folder (`dir`) and output folder (`savedir`) are defined and different, you can now keep the exact same name for the results files. This simplifies the steps in a complex workflow, where image analysis steps follow one another and file names are used as identifiers.

### Examples

Before:
``` bash
folder_A / X.tif
folder_B /
```

New commands:
```bash
$ cellpose --pretrained_model nuclei --dir ./folder_A/ --savedir ./folder_B/   # (1)
$ cellpose --pretrained_model nuclei --dir ./folder_A/ --savedir ./folder_B/ --output_name '_res'   # (2)
$ cellpose --pretrained_model nuclei --dir ./folder_A/ --savedir ./folder_B/ --output_name ''   # (3)

# (4)
$ cellpose --pretrained_model nuclei --dir ./folder_A/ --output_name ''
```

After new different commands:
``` bash
folder_B / X_cp_masks.tif # (1)
folder_B / X_res.tif # (2)
folder_B / X.tif # (3)

# (4) the user could rewrite the X file, but we prevent this by adding `_cp_masks`
folder_A / X.tif
folder_A / X_cp_masks.tif
```

### DIFF
--- a/cellpose/__main__.py
+++ b/cellpose/__main__.py
@@ -31,7 +31,7 @@ def main():
     """
 
     args = get_arg_parser().parse_args(
-    )  # this has to be in a seperate file for autodoc to work
+    )  # this has to be in a separate file for autodoc to work
 
     if args.version:
         print(version_str)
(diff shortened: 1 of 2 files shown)


carsen-stringer:
thanks, I updated to a samefile check to ensure that the directories are really not the same

PR #1114, opened 2025-02-21
Replace label loop with vectorized operation in _label_overlap

Tomvl117:
### Use `np.add.at` to more efficiently count overlapping labels in `_label_overlap`.

I had to make this adjustment for a separate project, in order to speed up processing of large datasets.

In essence, this change removes the need for a for-loop to compare overlapping labels. We take the label pairs `x` and `y`, and check whether they share any pixels. Then, we add 1 to `overlap` at the index positions that correspond to the label found in `x` and `y`. In other words, if `x[i] = A` and `y[i] = B`, we increment `overlap[A, B]` by 1.

As you can see, this is effectively the same operation as the for-loop, but it will be much faster for large datasets.

### DIFF
--- a/cellpose/metrics.py
+++ b/cellpose/metrics.py
@@ -156,12 +156,10 @@ def _label_overlap(x, y):
     # preallocate a "contact map" matrix
     overlap = np.zeros((1 + x.max(), 1 + y.max()), dtype=np.uint)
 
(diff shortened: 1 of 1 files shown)


Tomvl117:
This seems to be unnecessary with the Numba optimizations in place.

PR #1129, opened 2025-03-16
Report errors when saving masks to *_seg.npy

deroulers:
Currently, in the GUI, when saving the masks in a `*_seg.npy` file fails, no error message is printed and the process silently stops in the call to `np.save` in `gui/io.py`, function `_save_sets`.

With this PR, an error message would be displayed, giving hints about the cause of the failure (e.g. ``permission denied''), and the process would go further (e.g. updating the graphical display if needed, which should happen even if the call to `np.save` failed).


carsen-stringer:
thanks!

PR #1130, opened 2025-03-16
Enable to distinguish manual and automatic masks in GUI

deroulers:
With this PR, in the GUI, manually segmented mask are by default displayed differently than masks found by automatic segmentation: the transparency of the latter has a checkerboard pattern instead of being solid. This distinction can be toggled off and on by pressing 'M' (for manual) or (un)checking a box.


Jiadalee:
@carsen-stringer  is this feature pr merged to the recent version yet?

PR #1131, opened 2025-03-19
Compress TIFFs on save

qin-yu:
Hi @carsen-stringer, 

I made this small change because it saves a huge amount of storage in my projects. Segmentations compress really well, e.g. a 3D segmentation of `(263 x 1024 x 1024)` drops from 526 MB to 2.5 MB when using `zlib` with `tifffile.imwrite()`, about a 99.5% reduction.

I think most users saving segmentations with Cellpose would benefit from this, since the main bottleneck is usually I/O, not compression speed.

### DIFF
--- a/cellpose/gui/io.py
+++ b/cellpose/gui/io.py
@@ -591,6 +591,7 @@ def _save_flows(parent):
     else:
         print("ERROR: no flows or cellprob found")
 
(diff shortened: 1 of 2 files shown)


carsen-stringer:
thank you!

Issue #1162, opened 2025-05-08
[BUG] Cellpose SAM - NET ERROR: Numpy is not available


emiglietta:
**Describe the bug**
Running the CPSAM model on an RGB image from the GUI, I get this error:

```
2025-05-08 09:07:19,725 [INFO] WRITING LOG OUTPUT TO /Users/emigliet/.cellpose/run.log
2025-05-08 09:07:19,725 [INFO] 
cellpose version: 	4.0.2 
platform:       	darwin 
python version: 	3.10.17 
torch version:  	2.2.2
2025-05-08 09:07:20,202 [INFO] ** TORCH MPS version installed and working. **
GUI_INFO: loading image: /Users/emigliet/Documents/GitHub/emiglietta_projects/2025_Raspberry_Pollen_classification/Cellpose_pollen_segmentation/crops/15.10.48.3c.iii.59-1.tif
GUI_INFO: image shape: (512, 512, 3)
GUI_INFO: normalization checked: computing saturation levels (and optionally filtered image)
{'lowhigh': None, 'percentile': [1.0, 99.0], 'normalize': True, 'norm3D': True, 'sharpen_radius': 0.0, 'smooth_radius': 0.0, 'tile_norm_blocksize': 0.0, 'tile_norm_smooth3D': 0.0, 'invert': False}
[np.float32(0.0), np.float32(154.0)]
2025-05-08 09:07:47,774 [INFO] ** TORCH MPS version installed and working. **
2025-05-08 09:07:47,774 [INFO] >>>> using GPU (MPS)
2025-05-08 09:07:49,313 [INFO] >>>> loading model /Users/emigliet/.cellpose/models/cpsam
{'lowhigh': None, 'percentile': [1.0, 99.0], 'normalize': True, 'norm3D': True, 'sharpen_radius': 0.0, 'smooth_radius': 0.0, 'tile_norm_blocksize': 0.0, 'tile_norm_smooth3D': 0.0, 'invert': False}
NET ERROR: Numpy is not available
```

I'm tunning this on Mac M2 Pro with Sequoi 15.4.1. This is a fresh installation of Cellpose in a new conda environment (Python=3.10) using `python -m pip install 'cellpose[gui]'`, no other installations in this environment.

I used `pip freeze |grep numpy` to check numpy verion: `numpy==2.2.5`

Thank you for such a great tool, I'm eager to try the new version!!
Best!

mrariden:
[...]

PR #1168, opened 2025-05-12
Run transformer on MPS then do CPU post-processing

mrariden:
I forced self.device to be 'cpu' in `CellposeModel._compute_masks()` to switch to cpu post-processing after running the network on MPS. This works OK in the GUI since the GPU button is still checked, and the next time the button is pressed the GUI resets to start with MPS for the transformer.


mrariden:
resolves #1162

Issue #1169, opened 2025-05-13
[FEATURE] Support for AMD gpu's (and others) on windows


Teranis:
**Describe the solution you'd like**
Hi,
I'm working on Cell ACDC.
My personal computer runs Windows and has an AMD GPU. In order to run Cellpose on this GPU, I looked into DirectML, which uses Direct X to run PyTorch. Since DirectML doesn't support sparse tensors, I had to do some _slight_ operation of pytorch. I'll provide the code in additional context, where I move every operation and tensor which is not supported to the CPU.
Although I had my doubts about the performance, it still seems A LOT faster than just running it on the CPU. The only problem is that one needs to use python 3.11.
**Describe alternatives you've considered**
Waiting 5-10x longer while segmenting on CPU

**Additional context**
Code for making cellpose use the DirectML:
```
def setup_custom_device(model, device):
    model.gpu = True
    model.device = device
    model.mkldnn = False
    if hasattr(model, 'cp'):
        model.cp.gpu = True
        model.cp.device = device
        model.cp.mkldnn = False
        if hasattr(model.cp, 'net'):
            model.cp.net.to(device)
            model.cp.net.mkldnn = False
    if hasattr(model, 'net'):
        model.net.to(device)
        model.net.mkldnn = False
    if hasattr(model, 'sz'):
        model.sz.device = device

def setup_directML(model):
    print(
        'Using DirectML GPU for Cellpose model inference'
    )
    import torch_directml
    directml_device = torch_directml.device()
    setup_custom_device(model, directml_device)
```

Code for fixing sparse tensor implementation:
```
def fix_sparse_directML(verbose=True):
    """DirectML does not support sparse tensors, so we need to fallback to CPU
    """
    import torch
    import functools
    import warnings

    def fallback_to_cpu_on_sparse_error(func, verbose=True):
        @functools.wraps(func) # wrapper shinanigans (thanks chatgpt)
        def wrapper(*args, **kwargs):
            device_arg = kwargs.get('device', None)

            # Ensure indices are int64 if args[0] looks like indices
            if len(args) >= 1 and isinstance(args[0], torch.Tensor):
                if args[0].dtype != torch.int64:
                    args = (args[0].to(dtype=torch.int64),) + args[1:]

            try: # try to move result to dml
                result = func(*args, **kwargs)
                if device_arg is not None and str(device_arg).lower() == "dml":
                    try:
                        result.to("dml")
                    except RuntimeError as e:
                        if verbose:
[...]

PR #1181, opened 2025-05-16
DirectML support

Teranis:
See #1169  for details!

Adds functions to implement torch-directml. This library uses Directx12 to accelerate PyTorch, which is slower than CUDA but has the benefit of being compatible with any Directx12 device on Windows, including AMD GPUs.

Also includes working example!

### DIFF
--- a/cellpose/contrib/directml.py
+++ b/cellpose/contrib/directml.py
@@ -4,7 +4,7 @@
 # Limitations:
 # Officially support only up to PyTorch 2.4.1 (should be fine with cellpose)
 # Not yet out for python 3.13
-# Probably not the fastest option, but works suprisingly fast and was easy to implement
+# Probably not the fastest option, but works surprisingly fast and was easy to implement
 
 # Notes:
 # No additional drivers needed, but requires Windows 10/11 and a DirectX 12 compatible GPU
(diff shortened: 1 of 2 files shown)


mrariden:
@Teranis  This looks good! Can you also add some documentation in the `docs/installation.rst` file?

Teranis:
Hi, I modified the cellpose code so that it tries to use DirectML if CUDA and MPS are not available. I tried to follow the existing logic for it. I also included installation instructions (which are now a lot easier than before.

Teranis:
Hi @mrariden ,
I have added a unit test for `core.assign_device`. I'm not sure how I would realistically cover the other functionality. I would run everything, but with a different fixture as the model:
```
@pytest.fixture()
def cellposemodel_fixture_24layer_directml():
    """ This is functionally identical to CellposeModel but uses mock class """
    gpu = True
    with patch('torch.cuda.is_available', return_value=False), \
            patch('torch.backends.mps.is_available', return_value=False):
        model = MockCellposeModel(24, gpu=gpu)
    yield model
```
Then, I would compare the model output to the ground truth. I'm confused about how to do these last steps or why you strictly stick to CLI.

Kind regards,
Timon Stegmaier

PR #1183, opened 2025-05-16
Refactor image conversion functions for improved clarity and error ha…

mrariden:
better tests for transforms to catch things we missed previously. 

3d image reading fixes and explicit guessing about what the order is in the gui. API is more robust.


mrariden:
full test passing on ubuntu

mrariden:
gui appears to be working too

Guessing will fail for some images: 
https://github.com/MouseLand/cellpose/blob/9cb3c292a23068de91eadec3bb74bb6bfed7a523/cellpose/io.py#L236-266

PR #1193, opened 2025-05-20
Ansiotropy fix

### LINKED ISSUE #1191: [BUG] Anisotropy is ignored in CLI, creates an error in the GUI
If I enter a value >1 for the anisotropy flag in the CLI (for a zstack), cellpose still analyzes the image as if the zstack were isotropic (no resizing is done). When using the GUI, the analysis seems to treat the stack correctly and make it isotropic prior to segmentation, but then I get an opencv error related to resizing at the mask creation step. This is using v4.0.3. This is the error message:

`2025-05-20 08:41:31,452 [WARNING] Resizing is depricated in v4.0.1+
2025-05-20 08:41:32,596 [INFO] masks created in 5.33s
GUI_INFO: resizing flows to original image size
ERROR: OpenCV(4.11.0) D:\a\opencv-python\opencv-python\opencv\modules\imgproc\src\resize.cpp:3845: error: (-215:Assertion failed) !dsize.empty() in function 'cv::hal::resize'`

mrariden:
Anisotropy was running in the CellposeModel.eval() function but was not rescaled after segmentation resulting in incorrectly sized masks. Now, rescaling runs at the end of .eval if anisotropy isn't `None`. 

Also added passing test that failed with previous implementation. The tests require 3d so they only run on local with `--runslow` pytest flag. 

Fixes #1191

### DIFF
--- a/cellpose/models.py
+++ b/cellpose/models.py
@@ -339,8 +339,8 @@ class CellposeModel():
         
         masks, dP, cellprob = masks.squeeze(), dP.squeeze(), cellprob.squeeze()
 
-        # undo diameter resizing:
-        if image_scaling is not None:
+        # undo resizing:
+        if image_scaling is not None or anisotropy is not None:
             if do_3D:
                 # Rescale xy then xz:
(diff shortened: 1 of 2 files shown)


mrariden:
All tests passing on ubuntu

Issue #1197, opened 2025-05-21
TypeError: train_seg() got an unexpected keyword argument 'channels'


Jiadalee:
**Describe the bug**
I used CP3 a lot previously. In CP4, the train_seg() function seems not supporting the channels selection argument, but the CP3 does. See codes and error screenshot below. Removing the channels argument works fine.

**To Reproduce**
Steps to reproduce the behavior: simply run python codes below
```

```

**Run log**
Please post all command line/notebook output for us to understand the problem. For this please make sure you are running with verbose mode on. So command line, with `--verbose` tag, or in a notebook first run 
 Below are my codes:

```
# fine-tune cyto model (Note: cellpose model only accept image shape of [batch, channel, height, width] - channel first)
new_model_path, train_losses, val_losses = train.train_seg(model_pretrained.net,
                              train_data=train_images,
                              train_labels=train_masks,
                              test_data=val_images,
                              test_labels=val_masks,
                              channels=channels,
                              save_path=model_dir,
                              n_epochs=200,
                              learning_rate=0.01,
                              weight_decay=0.0001,
                              SGD=True,                  # set False to Use Adam optimizer
                              nimg_per_epoch=8,
                              model_name=model_name,
                              min_train_masks=0)         # Allow training with fewer masks
```

**Screenshots**
 see error mesage below

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[23], line 12
      9 model_name = "cyto3_finetuned_new"
     11 # fine-tune cyto model (Note: cellpose model only accept image shape of [batch, channel, height, width] - channel first)
---> 12 new_model_path, train_losses, val_losses = train.train_seg(model_pretrained.net,
     13                               train_data=train_images,
     14                               train_labels=train_masks,
     15                               test_data=val_images,
     16                               test_labels=val_masks,
     17                               channels=channels,
     18                               save_path=model_dir,
     19                               n_epochs=200,
     20                               learning_rate=0.01,
     21                               weight_decay=0.0001,
     22                               SGD=True,                  # set False to Use Adam optimizer
     23                               nimg_per_epoch=8,
     24                               model_name=model_name,
     25                               min_train_masks=0)         # Allow training with fewer masks
     27 # diameter of labels in training images
     28 diam_labels = model_pretrained.net.diam_labels.item()

TypeError: train_seg() got an unexpected keyword argument 'channels'
```

[...]

PR #1199, opened 2025-05-21
Fix IndexError in Segmentation Loss Function by Ensuring Proper Label Tensor Conversion

Jiadalee:
### Summary

This PR fixes an `IndexError` that occurred in the `_loss_fn_seg` function during model training. The issue was caused by improper indexing into the label array when it was still a NumPy array instead of a PyTorch tensor.

### Fixes
- Converts `lbl` to a PyTorch tensor if it's a NumPy array.
- Moves `lbl` to the appropriate device and casts to `float`.
- Adds shape validation to ensure the label has at least 3 channels.
- Adjusts indexing to safely extract `cellprob`, `flowY`, and `flowX`.

### Impact
This fix prevents training crashes due to label indexing and ensures consistent tensor formatting across training batches.

Tested and verified with:
- Input shapes: (N, 3, 128, 128)
- Original and fine-tuned Cellpose models

No changes to inference behavior.

### DIFF
--- a/cellpose/train.py
+++ b/cellpose/train.py
@@ -35,22 +35,34 @@ def _loss_fn_seg(lbl, y, device):
     Calculates the loss function between true labels lbl and prediction y.
 
     Args:
-        lbl (numpy.ndarray): True labels (cellprob, flowsY, flowsX).
+        lbl (np.ndarray or torch.Tensor): True labels (cellprob, flowsY, flowsX).
         y (torch.Tensor): Predicted values (flowsY, flowsX, cellprob).
         device (torch.device): Device on which the tensors are located.
 
     Returns:
         torch.Tensor: Loss value.
-
     """
+    if isinstance(lbl, np.ndarray):
+        lbl = torch.from_numpy(lbl)
+
+    lbl = lbl.to(device).float()  # Ensure it’s on the right device and float
+
+    if lbl.shape[1] < 3:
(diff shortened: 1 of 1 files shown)


mrariden:
Hi, can you describe the bug that you're trying to fix? Maybe opening an issue would better than a PR. 

The code changes you made don't agree with the first 3 bullets of your 'fixes' list. Is this still in progress?

Jiadalee:
> Hi, can you describe the bug that you're trying to fix? Maybe opening an issue would better than a PR.
> 
> The code changes you made don't agree with the first 3 bullets of your 'fixes' list. Is this still in progress?

Jiadalee:
@mrariden thank you for reviewing the pr. Per the root question/discussion from [#1197](https://github.com/MouseLand/cellpose/issues/1197). I need to manually select the channels for image data and remove the channel for mask data before feeding these datasets to the new CP4 model. Otherwise, the `indexError `in Segmentation Loss Function will pops up. If you checked my commits, I was also trying to revise the  `_loss_fn_seg convert`  function to dynamically ensure that the torch tensor variable  `y  (predicted masks)` to be numpy type so that both  the variable s ` lbl (true labels)` and  `y  (predicted masks)`  have consistent data type, either they are numpy or torch tensor data types. I finally I can manually make such data consistency without modifying the function of `_loss_fn_seg convert` . With that, we are good now. I will close this pr. Thank you again.

PR #1203, opened 2025-05-22
No masks fix

### LINKED ISSUE #1200: [BUG] running model with compute_masks=False causes cv2.resize error because masks=[0]
I just upgraded to cellpose 4.0.4 today. 

I ran the pretrained 'cyto' model with compute_masks=False, and a diameter value which was not `None`. This led to a cv2.resize error. 
Image

Tracing the error it occurs in `models.py` with the `model.eval`. When compute_masks=False, masks=[0]. However, it is the current code is still trying to resize this back to the input image resolution. Please could you fix this? I have drawn an arrow to indicate the line needing to fix - thanks !

Image

mrariden:
Attempted resizing on empty mask arrays failed when compute_masks was set to False in CellposeModel.eval(). Changed not resize masks if compute_masks was False, and kept returning empty array. 

fixes #1200


mrariden:
all tests passing on ubuntu

PR #1211, opened 2025-05-29
quick fix for the raised issue #12208 regarding the functions of io.imsave() and io.save_rois()

Jiadalee:
At  run_Cellpose-SAM.ipynb file, both io.imsave() and io.save_rois() should not return anything since the API specifies that these two functions return None.  Also, the `mask[i]`  is missing in the io.imsave() function. With that, I updated those codes to save images and rois in a proper way.


Jiadalee (commented):
pls skip the first four commits under this pr, since they are not useful. Pls only consider the commit of `Quick fix for issue1208 TypeError: imsave() and
 io.save_rois()`

Issue #1217, opened 2025-06-02
[BUG] Fine-tuning cp4 generates strange Training Loss


Jiadalee:
**Describe the bug**
I'm using the following codes and my own labels to fine-tune the cp4 model, but the train loss starts with a high value, then going down, but finally rise up a lot (see loss figure attached). Such a loss fluctuation pattern persist. Did I set any hyper-parameters incorrectly?

**To Reproduce**
My training and validation codes below:
```
try:
    model_pretrained = models.CellposeModel(gpu=True)
except Exception as e:
    print(f"An error occurred while loading the model: {e}")

#cp4: fine-tune cpsam model (Note: cpsam model only accepts image shape of [batch, height, width] - needs to extract channel)
model_name = "cpsam_finetuned_new"
new_model_path, train_losses, val_losses = train.train_seg(model_pretrained.net,
                              train_data=train_images_copy,
                              train_labels=train_masks_copy,
                              test_data=val_images_copy,
                              test_labels=val_masks_copy,
                              save_path=model_dir,
                              n_epochs=30,
                              learning_rate=0.01,
                              weight_decay=0.0001,
                              SGD=True,                  # set False to Use Adam optimizer
                              nimg_per_epoch=8,
                              model_name=model_name,
                              min_train_masks=0)         # Allow training with fewer masks
```

**Run log**

2025-06-02 09:45:49,614 [INFO] 
cellpose version: 	4.0.5.dev14+gc68d52e 
platform:       	linux 
python version: 	3.12.9 
torch version:  	2.6.0+cu124
2025-06-02 09:45:49,616 [WARNING] SGD is deprecated, using AdamW instead
2025-06-02 09:45:49,617 [INFO] computing flows for labels
100%|████████████████████████████████████████████████████████████████████████████████████| 99/99 [00:03<00:00, 29.05it/s]
2025-06-02 09:45:53,039 [INFO] computing flows for labels

100%|████████████████████████████████████████████████████████████████████████████████████| 11/11 [00:00<00:00, 28.45it/s]
2025-06-02 09:45:53,429 [INFO] >>> computing diameters

100%|██████████████████████████████████████████████████████████████████████████████████| 99/99 [00:00<00:00, 6176.35it/s]
100%|██████████████████████████████████████████████████████████████████████████████████| 11/11 [00:00<00:00, 4963.14it/s]
[...]

PR #1226, opened 2025-06-06
Fix test loss calculation for each epoch from issue #1217

Jiadalee:
The previous train_seg() function at the train.py calculates the test_loss at every 10 epochs plus the first epoch, which makes the test loss outputs strange to visualize. I've raised an issue #1217 at [link ](https://github.com/MouseLand/cellpose/issues/1217). For this PR, I made the following changes:

- Remove np.zeros(n_epochs) and replace with dynamic train_losses.append(...) and test_losses.append(...);
- Compute and store test loss at every epoch;
- Keep train_logger.info(...) frequency unchanged (only logs on epoch 5 or every 10 epochs); I initially changed this point but found it is not necessary at this moment;
- Add None to test_losses when no test data is provided.

Although test_loss is calculated at each epoch, there is no significant increase in computational cost if you do not set a too small batch size and also have decent GPU devices. This feature branch works fine to calculate test loss at each epoch, shown as below:

image

### DIFF
--- a/cellpose/train.py
+++ b/cellpose/train.py
@@ -425,19 +425,20 @@ def train_seg(net, train_data=None, train_labels=None, train_files=None,
     train_logger.info(f">>> saving model to {filename}")
 
     lavg, nsum = 0, 0
-    train_losses, test_losses = np.zeros(n_epochs), np.zeros(n_epochs)
+    train_losses, test_losses = [], []
+
     for iepoch in range(n_epochs):
         np.random.seed(iepoch)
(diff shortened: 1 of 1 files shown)


carsen-stringer:
Sorry computing the test loss at each epoch slows down training a bit, that's why we don't do it. Do you think you could make this as an option for the API but not the default?

Jiadalee:
> we've decided to keep the test losses as is for now, but please feel free to keep this code in your own fork of the repo, thanks!

ok, no problem

PR #1245, opened 2025-06-16
fix padding for smaller than 256 block size

mrariden:
The forward method is called for images with smaller than 256 size because of incorrect padding. Correct padding code to run the forward method less. 

Solution credit @carsen-stringer


carsen-stringer:
bug finding credit to @mrariden , thank you!

Issue #1255, opened 2025-06-26
[BUG] Cellpose-SAM/Cellpose4 not loading properly TCZYX images


bpavie:
**Describe the bug**
I try to process a 3d image in format TCZYX with cellpose-SAM via the CLI but loading failed

**To Reproduce**
I have an image of dimension TCZYX that I saved using Fiji Bi-Formats Exporter

I checked the dimension using bioio:

```python
from bioio import BioImage

img = BioImage("image.ome.tif")
print(img .dims)
print(img .data.shape)
print(img .dims.order)
```
```bash
<Dimensions [T: 1, C: 3, Z: 4, Y: 397, X: 397]>
(1, 3, 4, 397, 397)
TCZYX
```

Which indicate the the Z axis is 2 and C axis is 1
However, when I run the command line with the parameters `--z_axis 2` `--channel_axis 1` `--do_3D` it indicates that it is running on `running YX: 397 planes of size (4, 3)`

1. Run the command line
```bash
cellpose --image_path "image.ome.tif" --save_tif --z_axis 2 --channel_axis 1 --do_3D --verbose
```
2. See log
```bash
cellpose --image_path image.ome.tif" --save_tif --z_axis 2 --channel_axis 1 --do_3D --verbose
```

**Run log**
```
Welcome to CellposeSAM, cellpose v
cellpose version:       4.0.3
platform:               win32
python version:         3.12.0
torch version:          2.7.0+cu128! The neural network component of
CPSAM is much larger than in previous versions and CPU excution is slow.
We encourage users to use GPU/MPS if available.


2025-06-26 15:01:02,845 [INFO] WRITING LOG OUTPUT TO C:\Users\xxxxxx\.cellpose\run.log
2025-06-26 15:01:02,845 [INFO]
cellpose version:       4.0.3
platform:               win32
python version:         3.12.0
torch version:          2.7.0+cu128
2025-06-26 15:01:02,845 [INFO] >>>> using CPU
2025-06-26 15:01:02,846 [INFO] >>>> using CPU
2025-06-26 15:01:02,846 [WARNING] the '--no_resample' flag is deprecated in v4.0.1+ and no longer used
2025-06-26 15:01:02,846 [WARNING] the '--diam_mean' flag is deprecated in v4.0.1+ and no longer used
2025-06-26 15:01:02,850 [INFO] >>>> running cellpose on 1 images using all channels
[...]

PR #1258, opened 2025-06-27
Fix 3d reading cli

mrariden:
CLI 3d image reading would correctly read the image in the first time in `__main__` but would pass the `channel_axis` and `z_axis` to the `.eval()` function and then the image would get reshaped again sometimes, depending on the channel_axis. Needed to change the image reading function in `__main__` to just read the raw image and let the conversion function in `.eval` do its thing. 

All tests passing on ubuntu.


mrariden:
Resolves #1255

PR #1270, opened 2025-07-15
fix: remove black stripe artefacts when axis != -1 in

spkgyk:
The clean-up loop that sets un-normalised channels to 0 expected the channel dimension to be last.  However, since `np.moveaxis(..., -1, axis)` was done prior to this, instead of setting the channel to zero, it would set a row/column (depending on the axis) to zero instead, resulting in a black bar across the image.

### DIFF
--- a/cellpose/transforms.py
+++ b/cellpose/transforms.py
@@ -765,10 +765,6 @@ def normalize_img(img, normalize=True, norm3D=True, invert=False, lowhigh=None,
             transforms_logger.critical(error_message)
             raise ValueError(error_message)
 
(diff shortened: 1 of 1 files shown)


mrariden:
Looks good. What were the settings/image dimensions that produced the error?

spkgyk:
It happens when you pass in paths to the evaluation method. It loads the images, makes them channels first, and requests the normalisation function with axis=0. It happens whenever axis!=-1 I think.

spkgyk:
Will this be merged? Not trying to rush, just curious about the release schedule

PR #1311, opened 2025-08-22
Allow specify gpu device number

### LINKED ISSUE #1240: [BUG] Cellpose can only use gpu device 0
When I had 2 gpus available torch.cuda.device_count() == 2 and I tried to use gpu_device 1 when cuda:0 was already used - cellpose used the CPU instead of the GPU

The problem is that assign_device method never passes the device when it calls use_gpu(use_torch=True) and because of that use_gpu always tests device 0. So if GPU:0 is busy then assign_device will use CPU instead. The fix would be to call use_gpu(use_torch=use_torch,gpu_number=device)

mrariden:
fixes #1240. Passes device GPU string to `assign_device()` so it can actually set the device ID. 

- [x] Partial tests on Mac
- [x] Full tests on ubuntu 
- [x] passing GH workflow tests

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -6,23 +6,34 @@ labels: bug
 assignees: ''
 
 ---
+## Before you fill out this form:
+**Did you review the FAQ?**
+- [ ] Yes, I reviewed the FAQ on the [cellpose ReadTheDocs](https://cellpose.readthedocs.io/en/latest/faq.html) 
 
-**Describe the bug**
+**Did you look through previous (open AND closed) issues posted on GH?**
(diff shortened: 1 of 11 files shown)


cellpose/core.py
```diff
@@ -77,7 +77,7 @@ def assign_device(use_torch=True, gpu=False, device=0):
     if isinstance(device, str):
         if device != "mps" or not(gpu and torch.backends.mps.is_available()):
             device = int(device)
-    if gpu and use_gpu(use_torch=True):
+    if gpu and use_gpu(gpu_number=device, use_torch=True):
```
cgoina, line 80: IMO this should be use_gpu(gpu_number=device,use_torch=use_torch) to propagate all params passed to assign_device. If you don't care about use_torch why even have it as a parameter.

PR #1358, opened 2025-11-01
ImageJ ROIs have inscrutable names

MarinManuel:
I wanted to be able to compare measurements obtained using ImageJ and using a the _seg.npy directly from a script, but when I ask cellpose to generate ImageJ ROIs, they have uuid names that are very hard to read.

I modified the code to give the ROIs names that correspond to the labels of the masks.
I couldnt test it with a case where empty outlines are present, but hopefully the code works in that case too, and the correspondence is preserved.


cellpose/io.py
```diff
@@ -649,19 +649,16 @@ def save_rois(masks, file_name, multiprocessing=None):
...
     """
     outlines = utils.outlines_list(masks, multiprocessing=multiprocessing)
-    nonempty_outlines = [outline for outline in outlines if len(outline)!=0]
-    if len(outlines)!=len(nonempty_outlines):
-        print(f"empty outlines found, saving {len(nonempty_outlines)} ImageJ ROIs to .zip archive.")
-    rois = [ImagejRoi.frompoints(outline) for outline in nonempty_outlines]
-    file_name = os.path.splitext(file_name)[0] + '_rois.zip'
-
+    rois = []
+    for i,outline in enumerate(outlines):
```
mrariden: If there are outlines missing, it would be a bad idea to rely on enumerate to get the correct outline ID. I think you should check each outline id to be consistent with the masks.

cellpose/io.py
```diff
@@ -649,19 +649,16 @@ def save_rois(masks, file_name, multiprocessing=None):
...
+        if len(outline) > 0:
+            rois.append(ImagejRoi.frompoints(outline, name=str(i+1)))
 
-    # Delete file if it exists; the roifile lib appends to existing zip files.
-    # If the user removed a mask it will still be in the zip file
-    if os.path.exists(file_name):
-        os.remove(file_name)
```
mrariden: why was this removed?

mrariden (commented):
2 things to amend @MarinManuel.

cellpose/io.py
MarinManuel: You are right, I guess there could be a situation where the masks are not numbered sequentially. I think my new version takes care of that.

cellpose/io.py
MarinManuel: `roifile.roiwrite` takes a `mode=` argument that can be 'a' for append or 'w' for write. write already overwrites an existing file, so I thought that was simpler, but I can revert that part

mrariden: no this is ok

PR #1367, opened 2025-11-10
fix(utils): fixes a bug in fill_holes_and_remove_small_masks 

rubencardenes:
# Bug Fix: fill_holes_and_remove_small_masks - Incorrect Label Removal

## Problem Identified

When running celposeSAM with different cellprob_threshold I found that with lower thresholds I was missing cells, which was quite odd and should not be the case. The problem seems to be related the function `fill_holes_and_remove_small_masks` in `cellpose/utils.py`, that I think has a critical bug when removing small masks.

Here an example with two thresholds where some cells are missing for cellprob_threshold = -5, both using min_size=500 and flow_threshold=0.95

cellprob_threshold = 0 
results_inference_flowth0 95_cellprobth+0

cellprob_threshold = -5 (see that some obvious cells are missing)
results_inference_flowth0 95_cellprobth-5

The Buggy Code seems to be in (lines 645-647):
```python
counts = fastremap.unique(masks, return_counts=True)[1][1:]
masks = fastremap.mask(masks, np.nonzero(counts < min_size)[0] + 1)
fastremap.renumber(masks, in_place=True)
```

The code assumed mask labels are always sequential (1, 2, 3, ...), but this assumption is violated when `remove_bad_flow_masks()` in `dynamics.py` removes masks with poor flow quality.

**Example scenario:**
- After `get_masks_torch`: labels = [1, 2, 3, 4]
- After `remove_bad_flow_masks` removes mask 2: labels = [1, 3, 4] (non-sequential!)
- `fastremap.unique()` returns counts in order of labels: [count_1, count_3, count_4]
- Finding small mask at index 1 (label 3 is small)
- Buggy code: `index + 1 = 1 + 1 = 2` → tries to remove label 2 (which doesn't exist!)
- Should remove: label 3

## Impact

- **Wrong masks could be removed** when labels are non-sequential
- **Correct small masks might not be removed** if the wrong index is used
- This could lead to incorrect cell counts and missed small debris removal

## Proposed Fix

```python
uniq, counts = fastremap.unique(masks, return_counts=True)
# uniq[0] is background (0), so uniq[1:] are the actual mask labels
# counts[1:] are the corresponding counts
small_mask_indices = np.nonzero(counts[1:] < min_size)[0]
# Get the actual label values to remove (not indices)
labels_to_remove = uniq[1:][small_mask_indices]
masks = fastremap.mask(masks, labels_to_remove)
fastremap.renumber(masks, in_place=True)
```

This fix is applied to both filtering locations (before and after fill_voids).

Example image with cellprob_threshold = -5 after the fix 

results_inference_flowth0 95_cellprobth-5_fix

### DIFF
--- a/cellpose/gui/gui.py
+++ b/cellpose/gui/gui.py
@@ -1770,7 +1770,7 @@ class MainW(QMainWindow):
         if model_name is None or custom:
             self.get_model_path(custom=custom)
             if not os.path.exists(self.current_model_path):
-                raise ValueError("need to specify model (use dropdown)")
+                raise ValueError("Model file not found: need to specify model (use dropdown)")
 
         if model_name is None or not isinstance(model_name, str):
             self.model = models.CellposeModel(gpu=self.useGPU.isChecked(),
@@ -1867,7 +1867,7 @@ class MainW(QMainWindow):
                 self.logger.error("Flows don't exist, try running model again.")
                 return
             
-            maski = dynamics.resize_and_compute_masks(
+            maski = dynamics.compute_masks_and_clean(
                 dP=dP,
                 cellprob=cellprob,
                 niter=niter,
--- a/cellpose/vit_sam.py
+++ b/cellpose/vit_sam.py
@@ -81,9 +81,15 @@ class Transformer(nn.Module):
            
         return x1, torch.zeros((x.shape[0], 256), device=x.device)
     
-    def load_model(self, PATH, device, strict = False):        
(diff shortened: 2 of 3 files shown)


mrariden:
Look good @rubencardenes thanks for pointing this out. I added a test that caught the incorrect behavior before the fix, and passed after the fix was implemented. 

Will merge into main when tests are passing
- [ ] GH workflows (partial)
- [ ] Full tests on Ubuntu
- [ ] Mac (partial)

Issue #1300, opened 2025-08-08
[BUG] Cellpose model exported from 3.1.1.1 silently fails to work properly on latests 4.0 version


jianluz:
**Title**  
[BUG] Cellpose model exported from 3.1.1.1 silently fails to work properly on latest 4.0 version  

**Describe the bug**  
A custom, fine-tuned model (`cyto3_train0327`) exported from Cellpose version **3.1.1.1** does not work correctly when run with Cellpose **4.0.x** (tested on 4.0.6).  
The evaluation code runs without throwing any errors, but the output `masks` array is completely zero, resulting in a black image upon visualization.  
The same model and code produce correct segmentation results when run on Cellpose **3.1.1.1**, so the issue seems to be a compatibility problem between models exported from older versions and the latest 4.0 release.  

**To Reproduce**  
1. Export a fine-tuned model from Cellpose version 3.1.1.1 (example: `cyto3_train0327`).  
2. Install Cellpose 4.0.6.  
3. Run the following code:  
   ```python
   masks, flows, styles = cellpose_model.eval(frame_image, diameter=None, channels=[0, 0])
4. Observe that masks is all zeros (black when visualized), despite no errors being raised.

**Expected behavior**
The same segmentation results as in Cellpose 3.1.1.1, where the model correctly outputs masks with segmented cells.

**Run log**
In 4.0.6: no errors, masks all zeros.
In 3.1.1.1: works as expected.

To reproduce correct behavior, I had to downgrade to the specific commit of 3.1.1.1:

pip install git+https://github.com/mouseland/cellpose.git@fb22843e70d03f7884c301b7b72bedd7d9c3d2d9

phaleanopsis:
This is the same issue as [Issue 1216](https://github.com/MouseLand/cellpose/issues/1216). I think it would make sense to either let Cellpose3 models explicitly fail using Cellpose4 or documenting this behaviour more explicit, [for example in the docs where other issues of switching between cp3 and cp4 are documented](https://cellpose.readthedocs.io/en/latest/settings.html#settings) or in the README?

mrariden:
The latest release of cellpose (v3.1.1.3) is now explicitly not forwards compatible and gives this error

PR #1368, opened 2025-11-10
fail on loading cp3 model in cp4

mrariden:
Check the `state_dict` after loading to verify that it's not a CP3 model. This will need to be revised if we ever implement concurrent support for CP3 and CP4

Passing tests on:
- [x] GH partial tests
- [x] Ubuntu full tests
- [x] mac partial tests


mrariden:
resolves #1300

Issue #1375, opened 2025-11-27
Zarr 3 support for distributed


felixS27:
## Before you fill out this form:
**Did you review the FAQ?**
- [ x] Yes, I reviewed the FAQ on the [cellpose ReadTheDocs](https://cellpose.readthedocs.io/en/latest/faq.html) 

**Did you look through previous (open AND closed) issues posted on GH?**
- [ x] Yes, I [searched for related previous GH issues](https://github.com/MouseLand/cellpose/issues?q=is%3Aissue).

### **Is your feature request related to a problem? Please describe.**
Zarr upgraded their file storage format to a new version (v3) and changed some of their function interfaces (zarr.open does only accept zero or one positional argument). When using cellpose distributed one needs to install zarr <3 into the environment and save the images in zarr v2 format. In order to still be able to use cellpose distributed with v3 zarr arrays it would be nice to update the corresponding functions

### **Describe the solution you'd like**
A simple update of the functions interface relevant to zarr (which is just the zarr.open)

### **Describe alternatives you've considered**
-

### **Additional context**
I forked the repo and just changed one or two lines of codes.

<img width="1160" height="691" alt="Image" src="https://github.com/user-attachments/assets/7b1031da-a999-48eb-a62d-49d6c60a260f" />

Thanks you so much for the tool and all the hard work.

carsen-stringer:
Thanks @felixS27 , this is for @GFleishman to review when he has time, for now please use your branch

PR #1379, opened 2025-12-09
Fix to issue #1375 to make it also zarr v3 compatible

felixS27:
This is the fix to make the code for distributed cellpose also zarr v3 compatible.
As it is mentioned in issue #1375

### DIFF
--- a/cellpose/models.py
+++ b/cellpose/models.py
@@ -326,16 +326,37 @@ class CellposeModel():
             torch.cuda.empty_cache()
             gc.collect()
 
-        if resample:
+            if resample:
+                # upsample flows flows before computing them:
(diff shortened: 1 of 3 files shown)


mrariden:
The issue is that people still have zarr v2 arrays. Ideally, we'd have a way to support both. This solution doesn't appear to be backwards compatible. 

Although you'd have to juggle the installed library version, it would be better to have code that checks the current zarr version and then load according to the installed version. Also wrap the loading code with a helpful error if it fails.

felixS27:
Hi,
thank you for your answer. I just tried the fix with zarr version 2 and zarr version 3 and the fix works with both zarr versions.
The fix I propose is actually only to be more explicit in naming the input parameter of zarr.open() instead of relying on positional arguments. (As far as I understand in zarr 3 they are more strict about the input arguments and rely on named arguments)
So In my opinion there is no need to check the current zarr version.
But I am open to learn more.

PR #1403, opened 2026-02-09
Update return tuple in 'eval' function documentation

ralfox:
Parameter `diams` is not any more returned by function `eval`.

Removed `diams` from the return tuple in the function documentation.


mrariden:
thanks

PR #1408, opened 2026-02-23
add support for anisotropic values of the flow3D_smooth

yuriyzubov:
Currently `flow3D_smooth` only accepts a single integer, applying the same Gaussian smoothing sigma uniformly across all spatial axes (Z, Y, X).

If voxel spacing differs across axes, passing a vector value to the flow3D_smooth parameter (anisotropic smoothing) makes more sense. 

## Changes

- `flow3D_smooth` now accepts `int` or `list[int]` (length 3)
- If an `int` is passed, it is broadcast to `[val, val, val]` for backwards compatibility
- Updated docstring to reflect the new accepted types

## Example

```python
# Uniform smoothing
model.eval(img, do_3D=True, flow3D_smooth=2)

# Anisotropic smoothing
model.eval(img, do_3D=True, flow3D_smooth=[3, 0, 0])

### DIFF
--- a/cellpose/cli.py
+++ b/cellpose/cli.py
@@ -109,8 +109,8 @@ def get_arg_parser():
         "--min_size", required=False, default=15, type=int,
         help="minimum number of pixels per mask, can turn off with -1")
     algorithm_args.add_argument(
-        "--flow3D_smooth", required=False, default=0, type=float,
-        help="stddev of gaussian for smoothing of dP for dynamics in 3D, default of 0 means no smoothing")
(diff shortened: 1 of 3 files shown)


mrariden:
Hi @yuriyzubov please merge my `ring_artifacts-refactor` and this can be completed.

docs/do3d.rst
carsen-stringer: @yuriyzubov @mrariden the main reason to use flow3D_smooth is for the more cross-hatch like artifacts, I think those ring artifacts are rare - did you see them @yuriyzubov ?

PR #1418, opened 2026-03-20
3D GUI: compute ortho cell outlines separately

mrariden:
This PR improves the 3D and 2D GUI. 

1. Compute the outlines for cells in orthoviews separately so the outlines look more complete in the 3D GUI. Previously, the outlines were only computed in the XY plane for each Z slice. This produced ortho slices with incomplete outlines in the YZ/XZ views.  
2. Remove unused upsampling code (left over from CP3 upsampling models)
3. Fix `compute_restore()` not defined error when loading
4. Fix filter button and remove filter functionality

- [ ] GUI testing protocol
- [ ] tests passing


mrariden:
# Before fixing ortho outlines: 
image

# After fixing ortho outlines: 
image

mrariden:
This implementation isn't efficient enough for larger volumes.

Issue #1419, opened 2026-03-24
[FEATURE] Speed-up for multiple ROI deletion


chenyenchung:
## Before you fill out this form:
**Did you review the FAQ?**
- [x] Yes, I reviewed the FAQ on the [cellpose ReadTheDocs](https://cellpose.readthedocs.io/en/latest/faq.html) 

**Did you look through previous (open AND closed) issues posted on GH?**
- [x] Yes, I [searched for related previous GH issues](https://github.com/MouseLand/cellpose/issues?q=is%3Aissue).

## Now fill this form out completely:

Deleting several masks from the GUI currently removes them one by one and reindexes the full mask state after each deletion. This preserves correctness, but it scales poorly for region-delete and multi-select workflows, especially on images with many ROIs or 3D stacks. In larger multi-delete operations, the GUI can appear unresponsive while the deletions are processed.

It seems feasible to delete multiple ROIs first and reindex once at the end. I tried a small proof-of-concept implementation that appears to behave correctly and improves efficiency for this use case. I plan to open a PR so you can decide whether this approach is appropriate for upstream.

carsen-stringer:
Thanks @chenyenchung , @mrariden will work on this when he has time

PR #1420, opened 2026-03-24
Speed up multi-ROI deletion in the GUI

chenyenchung:
This a proof-of-concept change associated with #1419 .

- Add batch deletion utilities in `cellpose/gui/delete_utils.py` to normalize ROI IDs and relabel masks in a single pass.
- Update `cellpose/gui/gui.py` so multi-ROI deletion uses the batch path, while single-ROI deletion remains the same.
- Preserve existing GUI-side effects, including ROI count updates, saved-set updates, and deletion tracking.
- Add regression tests in `tests/test_gui_delete_utils.py` to verify that the batch path behaves in the same way as the original deletion behavior in both 2D and 3D cases.
- Treat invalid or duplicate ROI IDs as no-ops.

## Testing
- Added a fixed-case regression test.
- Added randomized 2D and 3D equivalence tests against the legacy deletion path.
- Added a no-op test for invalid ROI IDs.

### DIFF
--- a/cellpose/gui/gui.py
+++ b/cellpose/gui/gui.py
@@ -16,7 +16,7 @@ import numpy as np
 from scipy.stats import mode
 import cv2
 
-from . import guiparts, menus, io
+from . import guiparts, menus, io, delete_utils
 from .. import models, core, dynamics, version, train
 from ..utils import download_url_to_file, masks_to_outlines, diameters
 from ..io import get_image_files, imsave, imread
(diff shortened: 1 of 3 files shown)


cellpose/gui/gui.py
```diff
@@ -62,65 +77,796 @@ def make_bwr():
...
+            197,
+            205,
+            213,
+            222,
+            230,
+            238,
+            246,
+            254,
+        ]
+    )
```
mrariden: @chenyenchung please undo the formatter-related changes

chenyenchung: My apologies for the noise. I should have disabled my auto-formatter. I’ve rebased and force-pushed to revert those changes and keep the diff focused on the actual logic. Ready for review again.

chenyenchung:
Rebased the commit to the current main.

PR #1428, opened 2026-04-03
GUI fixes: Add escape drawing, disable right-click drag, disable double left-click

mrariden:
- Add escape drawing by hitting escape key. 
- Disable unused right-click and drag to zoom behavior. This is not needed because other zoom controls exist and right-click to drag occasionally gets in the way of annotation. 
- Disable double left-click to reset the view. This was especially painful when looking at orthoviews in the 3D gui and quickly clicking around to see the orthoviews update. 

- [x] Update docs to describe escape draw behavior
- [x] GUI testing protocol
- [x] tests passing

### DIFF
--- a/cellpose/gui/gui.py
+++ b/cellpose/gui/gui.py
@@ -1018,7 +1018,7 @@ class MainW(QMainWindow):
         self.remove_roi_obj = None
 
     @property
-    def color(self):
+    def color(self) -> str:
         """Current color display mode as a lowercase string.
 
         Reflects the current selection of the RGBDropDown widget. Possible
--- a/cellpose/gui/gui.py
+++ b/cellpose/gui/gui.py
@@ -730,6 +730,11 @@ class MainW(QMainWindow):
                     self.brush_choose()
                 if not updated:
                     self.update_plot()
+            
+            # when in stroke, allow escaping out of drawing
+            else: 
+                if event.key() == QtCore.Qt.Key_Escape:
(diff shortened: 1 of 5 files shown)


mrariden:
~~This would be a good opportunity to de-duplicate the `keyPressEvent` code that is present in both `gui.py` and `gui3d.py`. I would pass the events that can be handled by the parent class up the hierarchy, so the 3D key press code is specific to the 3D GUI.~~ 

Done in another PR

mrariden:
# Tests passing on 
- [x] Mac
- [ ] Ubuntu 


# GUI manual testing: 
## Image loading
- [x] Drop image into GUI
- [x] Load from menu

## Network
    - [x] Diameter
    - [x] Flow threshold
    - [x] niter dynamics
- [x] User models work

## GUI: display
- [ ]  Auto adjust saturation checkbox and sliders work

## GUI: mouse/ROIs
- [x] Scroll to zoom
- [x] Esc to escape drawing

## GUI: menus
- [ ]  Models >
    - [ ] Add model and remove model works

## Filtering
- [ ]  Smoothing in 3D works

mrariden:
GUI status is good for merging, there are unrelated issues that should be fixed in other PRs:

1. Filter button doesn't work when you click it twice. You need to remove the filter and then apply it with different settings. 
2. Click 'auto adjust' saturation doesn't immediately snap to auto adjust.
3. User trained models are incorrect in drop down

mrariden:
Still need to replace the reset view functionality with a new key

PR #1432, opened 2026-04-09
Remove `.views` & `.color` states, fix gradZ 3D GUI display

mrariden:
Moved the main gui's `.view` and `.color` attributes to a stateless property that just looks at the actual drop down state. Setting the attribute/property is backwards compatible to use indexes or strings for convenience.  

Also, using more intuitive variables for the update_plot function. 

Also, removed duplicated code from 3d GUI `keyPressEvent` that is now delegated to `MainW` class. 

Fix gradZ displaying in 3d GUI. 

- [x] GUI testing protocol 
- [x] passing CI tests

### DIFF
--- a/cellpose/gui/gui.py
+++ b/cellpose/gui/gui.py
@@ -1425,7 +1425,7 @@ class MainW(QMainWindow):
         flowp_map = {
             'gradXY' : 0,
             'cellprob' : 1,
-            'gradZ' : 4,
+            'gradZ' : 2,
         }
         rgb_list = ['red', 'green', 'blue']
 
--- a/cellpose/gui/gui.py
+++ b/cellpose/gui/gui.py
@@ -31,15 +31,6 @@ except:
 
 Horizontal = QtCore.Qt.Orientation.Horizontal
 
-from contextlib import contextmanager
-
-@contextmanager
-def suppress_signals(widget):
-    widget.blockSignals(True)
-    try:
-        yield
-    finally:
-        widget.blockSignals(False)
 
 class Slider(QRangeSlider):
 
@@ -1008,9 +999,9 @@ class MainW(QMainWindow):
         self.ismanual = np.zeros(0, "bool")
 
         # -- set menus to default -- #
-        with suppress_signals(self.RGBDropDown):
+        with QtCore.QSignalBlocker(self.RGBDropDown):
             self.color = 'RGB'
-        with suppress_signals(self.ViewDropDown):
+        with QtCore.QSignalBlocker(self.ViewDropDown):
             self.view = 'image'
         self.delete_restore()
 
@@ -1517,7 +1508,7 @@ class MainW(QMainWindow):
         for r in range(3):
         # setValue on the slider triggers update_plot() so it needs to be suppressed
             slider = self.sliders[r]
-            with suppress_signals(slider):
+            with QtCore.QSignalBlocker(slider):
                 slider.setValue([
                     self.saturation[r][self.currentZ][0],
                     self.saturation[r][self.currentZ][1]
--- a/cellpose/gui/gui.py
+++ b/cellpose/gui/gui.py
@@ -811,8 +811,6 @@ class MainW(QMainWindow):
             self.sliders[n].setEnabled(True)
 
         self.toggle_mask_ops()
-
-        self.update_plot()
         self.setWindowTitle(self.filename)
 
     def disable_buttons_removeROIs(self):
--- a/cellpose/gui/gui.py
+++ b/cellpose/gui/gui.py
@@ -1454,7 +1454,7 @@ class MainW(QMainWindow):
         flowp_map = {
             'gradXY' : 0,
             'cellprob' : 1,
-            'gradZ' : 2,
+            'gradZ' : 4,
         }
         rgb_list = ['red', 'green', 'blue']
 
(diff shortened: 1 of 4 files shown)


mrariden:
# Tests passing on 
- [x] Mac
- [x] Ubuntu 


# GUI manual testing: 
## Image loading
- [x] Drop image into GUI
- [x] Load from menu

## Network
    - [x] Diameter
    - [x] Flow threshold
    - [x] niter dynamics
- [x] User models work

## GUI: display
- [x]  Auto adjust saturation checkbox and sliders work

## GUI: mouse/ROIs
- [x] Scroll to zoom

## GUI: menus
- [x]  Models >
    - [x] Add model and remove model works

## Filtering
- [x]  Smoothing in 3D works

mrariden:
Verified that `update_plot()` isn't called too frequently, only 1 or 2 times depending on the data update.

mrariden:
Verified that update plot is only called once per interaction :+1: 

Good to merge

cellpose/gui/gui.py
```diff
@@ -328,14 +328,24 @@ def make_buttons(self):
         self.l0.addWidget(self.satBox, b, 0, 1, 9)
 
         widget_row = 0
-        self.view = 0  # 0=image, 1=flowsXY, 2=flowsZ, 3=cellprob
-        self.color = 0  # 0=RGB, 1=gray, 2=R, 3=G, 4=B
```
mrariden, line 332: removes the .view and .color state so we only reference the checkbox state. This way we can use qt signals from the checkbox easily

cellpose/gui/gui.py
```diff
@@ -328,14 +328,24 @@ def make_buttons(self):
...
         self.RGBDropDown.addItems(
             ["RGB", "red=R", "green=G", "blue=B", "gray", "spectral"])
+        self.RGBDropDown.name_map = {
+            "rgb" : "rgb",
+            "red" : "red=r", 
+            "green" : "green=g", 
+            "blue" : "blue=b", 
+            "gray" : "gray", 
+            "spectral": "spectral",
+        }
```
mrariden, line 344: convience maps to get the correct drop down setting. These can be extended

cellpose/gui/gui.py
```diff
@@ -662,62 +673,75 @@ def level_change(self, r):
             self.update_plot()
 
     def keyPressEvent(self, event):
```
mrariden, line 675: modified this method to use qt events and event propagation instead of `updated` flag

cellpose/gui/gui.py
```diff
@@ -728,8 +752,6 @@ def keyPressEvent(self, event):
                         gci = min(count - 1, gci + 1)
                     self.BrushChoose.setCurrentIndex(gci)
                     self.brush_choose()
-                if not updated:
-                    self.update_plot()
```
mrariden, line 732: removed unnecessary update. use the event accepted flag instead

cellpose/gui/gui.py
```diff
@@ -789,8 +811,6 @@ def enable_buttons(self):
             self.sliders[n].setEnabled(True)
 
         self.toggle_mask_ops()
-
-        self.update_plot()
```
mrariden, line 793: removed unnecessary update. only really matters at startup

cellpose/gui/gui.py
```diff
@@ -913,7 +933,6 @@ def toggle_masks(self):
             self.draw_layer()
             self.update_layer()
         if self.loaded:
-            self.update_plot()
```
mrariden, line 916: removed unnecessary update. Toggling the checkbox is wired to `update_plot` via a signal

cellpose/gui/gui.py
```diff
@@ -1011,9 +1112,9 @@ def delete_restore(self):
     def clear_restore(self):
         """ delete restored imgs and reset settings """
         print("GUI_INFO: clearing restored image")
-        self.ViewDropDown.model().item(self.ViewDropDown.count() - 1).setEnabled(False)
-        if self.ViewDropDown.currentIndex() == self.ViewDropDown.count() - 1:
-            self.ViewDropDown.setCurrentIndex(0)
+        self.enable_restored_view(False)
+        if self.view == 'restored': 
+            self.view = 'image'
```
mrariden, line 1117: a good example of the gained clarity when using the new properties

cellpose/gui/gui.py
```diff
@@ -1381,26 +1493,26 @@ def update_plot(self):
...
+            if len(self.flows[flowp_map[self.view]]) > 0:
+                image = self.flows[flowp_map[self.view]][self.currentZ]
+            self.img.setImage(image, autoLevels=False, lut=self.bwr)
             self.img.setLevels([0.0, 255.0])
 
         for r in range(3):
-            self.sliders[r].setValue([
-                self.saturation[r][self.currentZ][0],
-                self.saturation[r][self.currentZ][1]
-            ])
```
mrariden, line 1403: the sliders were wired to fire a signal that connected to `update_plot()` previously. Now, this signal is blocked

cellpose/gui/gui.py
```diff
@@ -1972,18 +2081,17 @@ def compute_segmentation(self, custom=False, model_name=None, load_model=True):
                 Lz, Ly, Lx = self.NZ, self.Ly, self.Lx
                 Lz0, Ly0, Lx0 = flows_new[0].shape[:3]
                 print("GUI_INFO: resizing flows to original image size")
-                for j in range(len(flows_new)):
-                    flow0 = flows_new[j]
+                for flows0 in flows_new:
```
mrariden, line 2084: just a simplification/clarification

cellpose/gui/gui.py
```diff
@@ -2009,3 +2117,28 @@ def compute_segmentation(self, custom=False, model_name=None, load_model=True):
...
+        idx = dropdown.currentIndex() + increment
+
+        for _ in range(num_items):
+            idx %= num_items
+            if enabled[idx]:
+                dropdown.setCurrentIndex(idx)
+                return
+            idx += increment
+
+        self.logger.error('Could not find an emabled dropdown item.')
```
mrariden, line 2144: allows indexing a dropdown by `increment` and skips disabled items 

is used for RGB and views dropdowns

cellpose/gui/gui3d.py
```diff
@@ -576,89 +587,48 @@ def update_plot(self):
...
+                        event.accept()
+                        return
         if event.key() == QtCore.Qt.Key_Minus or event.key() == QtCore.Qt.Key_Equal:
             self.p0.keyPressEvent(event)
+            event.accept()
+            return
+
+        # propagate unhandled events to MainW                                                                                   
+        if not event.isAccepted():
+            super().keyPressEvent(event)
```
mrariden, line 631: deduplicated key press logic in gui3d so 2d gui takes care of it

PR #1438, opened 2026-04-28
Resample fix

mrariden:
@jeskowagner found that masks weren't being correctly resized and proposed a fix in #1434. This fix is preferred to be more consisted with the CP3 code. 

Added a test that fails to highlight the issue, and adding the fix causes the test to pass.

### DIFF
--- a/cellpose/dynamics.py
+++ b/cellpose/dynamics.py
@@ -2,6 +2,8 @@
 Copyright © 2025 Howard Hughes Medical Institute, Authored by Carsen Stringer , Michael Rariden and Marius Pachitariu.
 """
 import os
+import cv2
+from cellpose import transforms
 from scipy.ndimage import find_objects, center_of_mass, mean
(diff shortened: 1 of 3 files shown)


cellpose/dynamics.py
```diff
@@ -614,7 +616,22 @@ def resize_and_compute_masks(dP, cellprob, niter=200, cellprob_threshold=0.0,
                             device=device)
 
     if resize is not None:
-        dynamics_logger.warning("Resizing is deprecated in v4.0.1+")
+        if len(resize) == 2:
```
mrariden, line 619: This block was removed when I purged all resample/rescale code from CP4. It's verbatim cp3, and now actually does what the function name implies.

mrariden:
Passing tests: 
- [X] Ubuntu 
- [x] Mac
- [x] GH Workflows

PR #1441, opened 2026-05-01
Fix npy channel

mrariden:
Changing the `MainW.color` from `int` to `str` caused loading/saving .npy files to fail when loading `'current_channel'`. Added backwards-compatible check that allows old .npy files to be opened with new cellpose versions. In that case, the behavior is to set default `'current_channel'` to RGB. 

Also ensured GUI loads image with `update_plot()` when loading .npy file. 

- [ ] GUI testing protocol
- [ ] GH tests
- [x] Mac/ubuntu tests

### DIFF
--- a/cellpose/gui/io.py
+++ b/cellpose/gui/io.py
@@ -329,7 +329,13 @@ def _load_seg(parent, filename=None, image=None, image_file=None, load_3D=False)
             parent.ismanual = dat["ismanual"]
 
     if "current_channel" in dat:
-        parent.color = (dat["current_channel"] + 2) % 5
+        color = dat['current_channel']
+
+        # color updated to strings, force int channels to rgb for backwards compatibility
+        if isinstance(color, int):
+            parent.color = 'rgb'
+        elif isinstance(color, str):
+            parent.color = color
 
     if "flows" in dat:
         parent.flows = dat["flows"]
@@ -559,7 +565,7 @@ def _save_sets(parent):
                 parent.cellcolors[1:],
             "masks":
                 parent.cellpix,
-            "current_channel": (parent.color - 2) % 5,
+            "current_channel": parent.color,
             "filename":
                 parent.filename,
             "flows":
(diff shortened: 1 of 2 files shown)


carsen-stringer:
Thanks Michael, so seg.npy files will still open the same exact way as before?

cellpose/gui/io.py
```diff
@@ -329,7 +329,14 @@ def _load_seg(parent, filename=None, image=None, image_file=None, load_3D=False)
             parent.ismanual = dat["ismanual"]
 
     if "current_channel" in dat:
-        parent.color = (dat["current_channel"] + 2) % 5
+        color = dat['current_channel']
+
+        # old .npy files stored channel as int; default to 'rgb'
+        if isinstance(color, int):
```
mrariden: Backwards compatibility forces 'rgb' color when opening old _seg.npy files

mrariden: This can (should) be updated to convert the `int` to the correct `str`

mrariden:
use this script to fix intermediate `_seg.npy` files: 

```python 
import click
import numpy as np
from pathlib import Path


@click.command()
@click.argument('filename')
def main(filename: str|Path):
    filename = Path(filename)

    file_data = np.load(filename, allow_pickle=True).item()
    color = file_data['current_channel']
    if isinstance(color, str):
        click.echo(f'found incorrect `color` type: ({str(color)}), changing to int')
        file_data['current_channel'] = 0
    else:
        click.echo('file seems okay, exiting...')
        return 
    
    new_fn = filename.resolve().with_name(filename.stem[:-4] + '_fixed_seg.npy')
    click.echo(f'{">" * 5} Saving fixed file to \t{str(new_fn)}')
    np.save(new_fn, file_data)
    click.echo('done.')
    click.echo('You will have to rename the `_fixed_seg.npy` to end in `_seg.npy` (remove `_fixed_seg.npy`)')


if __name__ == "__main__":
    main()
```

mrariden:
Now the `.color` attribute is always saved as an int. All behavior is backwards compatible

Issue #1435, opened 2026-04-20
[FEATURE] discussion with dynamic efficiency improvement


weiyusheng:
English is not my native language, please excuse my spell errors.

# problem

I found it slow with cpsam model.

So I run this model with:

```
with profile(activities=[ProfilerActivity.CPU, ProfilerActivity.CUDA], record_shapes=True) as prof:
    with torch.inference_mode():
        with torch.autocast(device_type='cuda', dtype=torch.float16):
            masks, flows, styles = model.eval(...)
print(prof.key_averages().table(sort_by="cuda_time_total", row_limit=10))
```

env:

```
cpu i7-11700
gpu RTX 4070
gpu driver 595.97
cuda 13.2
2d (2k,2k) image
fp16 eval
```

And I find that plenty of time is cost during Aten::_copy both on CPU and GPU.

--------------------------------------------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  
                                              Name    Self CPU %      Self CPU   CPU total %     CPU total  CPU time avg     Self CUDA   Self CUDA %    CUDA total  CUDA time avg    # of Calls
                                       aten::copy_        26.98%        3.332s        27.06%        3.342s     146.784us        2.059s        14.12%        2.069s      90.865us         22766
                                         aten::add         2.93%     361.768ms         3.36%     414.731ms      18.883us        1.707s        11.71%        1.763s      80.263us         21963
                                      aten::select        11.11%        1.373s        11.40%        1.408s      16.451us        1.109s         7.60%        1.694s      19.796us         85569
                                          aten::mm         0.66%      82.099ms         0.66%      82.099ms      77.745us     894.875ms         6.14%     894.875ms     847.420us          1056
                                       aten::copy_        26.98%        3.332s        27.06%        3.342s     146.784us        2.059s        14.12%        2.069s      90.865us         22766
                                         aten::add         2.93%     361.768ms         3.36%     414.731ms      18.883us        1.707s        11.71%        1.763s      80.263us         21963
                                      aten::select        11.11%        1.373s        11.40%        1.408s      16.451us        1.109s         7.60%        1.694s      19.796us         85569
                                       aten::copy_        26.98%        3.332s        27.06%        3.342s     146.784us        2.059s        14.12%        2.069s      90.865us         22766
                                         aten::add         2.93%     361.768ms         3.36%     414.731ms      18.883us        1.707s        11.71%        1.763s      80.263us         21963
[...]

PR #1447, opened 2026-05-06
speed up w/ flattened inds instead of tuple (claude suggestion) and #…

carsen-stringer:
…#1435 suggestion


carsen-stringer:
I need to fix for resample

PR #1448, opened 2026-05-07
allow errors to propagate past Qt, add debug logging

mrariden:
Qt silences exceptions originating from built in methods, (e.g. `hoverEvent`, `keyPressEvent`, etc) making debugging difficult. This PR adds a function decorator that wraps these problematic functions to guarantee exception reporting. 

Current behavior reports that an error occurs to the user and the full stack trace is saved to the log in `~/.cellpose/run.log`

### DIFF
--- a/cellpose/gui/guiparts.py
+++ b/cellpose/gui/guiparts.py
@@ -1,6 +1,8 @@
 """
 Copyright © 2025 Howard Hughes Medical Institute, Authored by Carsen Stringer , Michael Rariden and Marius Pachitariu.
 """
+import logging
+import traceback
 from qtpy import QtGui, QtCore
 from qtpy.QtGui import QPixmap, QDoubleValidator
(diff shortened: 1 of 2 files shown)


cellpose/gui/guiparts.py
```diff
@@ -629,58 +642,56 @@ def __init__(self, image=None, viewbox=None, parent=None, **kargs):
...
+            y, x = int(ev.pos().y()), int(ev.pos().x())
+            if y >= 0 and y < self.parent.Ly and x >= 0 and x < self.parent.Lx:
+                if ev.button() == QtCore.Qt.LeftButton and not ev.double():
+                    idx = self.parent.cellpix[self.parent.currentZ][y, x]
+                    if idx > 0:
+                        if ev.modifiers() & QtCore.Qt.ControlModifier:
+                            # delete mask selected
+                            self.parent.remove_cell(idx)
+                        elif ev.modifiers() & QtCore.Qt.AltModifier:
+                            self.parent.merge_cells(idx)
```
mrariden, line 675: Just de-indented the code's try/catch to use the function decorator instead

PR #1451, opened 2026-05-12
Escape drawing, move reset view to space, disable right click zoom

### LINKED ISSUE #1445: [BUG] Drawing breaks due to `cv2.findContours` handling
When drawing, sometimes multiple contours are found for a single mask, but the code expects to encounter only a single contour. 

For example: 
https://github.com/MouseLand/cellpose/blob/94ae120369bdaea12da4921b81bc1fc7c8360558/cellpose/gui/gui.py#L1574-L1575

and 

https://github.com/MouseLand/cellpose/blob/94ae120369bdaea12da4921b81bc1fc7c8360558/cellpose/gui/gui.py#L1589-L1591

and in 3d GUI: 
https://github.com/MouseLand/cellpose/blob/94ae120369bdaea12da4921b81bc1fc7c8360558/cellpose/gui/gui3d.py#L292-L294

and here:
https://github.com/MouseLand/cellpose/blob/94ae120369bdaea12da4921b81bc1fc7c8360558/cellpose/gui/gui3d.py#L308-L310

and also in `merge_cells()`:

https://github.com/MouseLand/cellpose/blob/94ae120369bdaea12da4921b81bc1fc7c8360558/cellpose/gui/gui.py#L1318-L1320

This error manifests when there is a small hole in a mask that is being overlapped that is right on the edge of a mask, (see image below). Presumabely there are also other edge cases when this occurs. No errors are ever seen because the `add_mask()` functions are called in the overriden Qt virtual methods (`keyPressEvent`, `mouseClickEvent`) which is, apparently [by design](https://stackoverflow.com/questions/45787237/exception-handled-surprisingly-in-pyqt-pyside-slots). 

Example breaking mask (mask is on the left with a single missing interior pixel):
Image

Two changes should be made: 
1. Handle the edge case for when multiple contours are found.
2. Implement more robust drawing logic that will allow drawing to fail gracefully.

mrariden:
This is a rework of #1428

### Added/modified functionality
- Add escape drawing by hitting escape key.
- Disable unused right-click and drag to zoom behavior. This is not needed because other zoom controls exist and right-click to drag occasionally gets in the way of annotation.
- Disable double left-click to reset the view. This was especially painful when looking at orthoviews in the 3D gui and quickly clicking around to see the orthoviews update. Resetting the view moved to space key

### Bugfix
- In some cases, drawing would produce multiple contours. This wasn't explicitly checked before. Now the largest contour is always extracted from the contour list. Fixes #1445

### Items to complete before merging
- [x] Update docs to describe escape draw behavior
- [x] GUI testing protocol
- [x] tests passing
  - [x] Mac
  - [x] Ubuntu
  - [ ] GH workflow


mrariden:
# Tests passing on 
- [x] Mac
- [ ] Ubuntu 


# GUI manual testing: 
## Image loading
- [x] Drop image into GUI
- [x] Load from menu

## Network
    - [x] Diameter
    - [x] Flow threshold
    - [x] niter dynamics
- [x] User models work

## GUI: display
- [x]  Auto adjust saturation checkbox and sliders work
- [x] Space bar reset view

## GUI: mouse/ROIs
- [x] Scroll to zoom
- [x] Escape drawing with esc

## GUI: menus
- [x]  Models >
    - [x] Add model and remove model works

## Filtering
- [x]  Smoothing in 3D works

PR #1453, opened 2026-05-14
Fix IndexError when loading image with auto-adjust saturation unchecked

chenyenchung:
Dear Cellpose team,

Thank you for the great tool! Please find below a bug report and a proposed fix for saturation updates in the GUI:

In Zstack mode (`python -m cellpose --Zstack`), if an image is loaded when auto-adjust saturation is unchecked, an `IndexError: list index out of range is raised`:

```
File ".../cellpose/gui/gui3d.py", line 348, in move_in_Z
    self.update_plot()
File ".../cellpose/gui/gui3d.py", line 568, in update_plot
    super().update_plot()
File ".../cellpose/gui/gui.py", line 1372, in update_plot
    self.img.setLevels(self.saturation[self.color - 1][self.currentZ])
IndexError: list index out of range
```

This appears to be introduced in a982535 when saturation update when `parent.NZ` changes was disabled. This causes self.saturation not updating its shape at loading and remaining `[]` or per the last image loaded when auto-adjust saturation was checked. The first time `update_plot()` indexes `self.saturation[c][self.currentZ]` (e.g. via `move_in_Z` after the user scrolls, or immediately during the initial render at `currentZ = NZ // 2`), the list is shorter than the new `NZ` and indexing fails.

To fix this issue, in `cellpose/gui/io.py` `_initialize_images`, after `parent.NZ` is updated to the new image's Z depth, add an else branch to the existing if `parent.autobtn.isChecked():` block that re-sizes self.saturation to 3 × `NZ`. The user's previous per-channel values are preserved by broadcasting each channel's first-Z value across the new Z range; channels with no previous value default to [0, 255].

The structure is the same 3 × NZ × 2 (channel × Z × [lo, hi]) that `compute_saturation` produces, so `update_plot`, `level_change`, and the ortho views can index it without further changes.

A new test asserting `self.saturation` is 3 × NZ after loading with auto-adjust off is also added.

Please let me know if you need more information. Thank you.

Best,
Yen

### DIFF
--- a/cellpose/gui/gui.py
+++ b/cellpose/gui/gui.py
@@ -177,7 +177,7 @@ class MainW(QMainWindow):
     def __init__(self, image=None, logger=None):
         super(MainW, self).__init__()
 
-        self.logger = logger
+        self.logger = logger if logger is not None else logging.getLogger(__name__)
         pg.setConfigOptions(imageAxisOrder="row-major")
         self.setGeometry(50, 50, 1200, 1000)
         self.setWindowTitle(f"cellpose v{version}")
--- a/tests/test_gui.py
+++ b/tests/test_gui.py
@@ -2,7 +2,6 @@ import numpy as np
 import pytest
 
 
-@pytest.mark.slow
 @pytest.fixture(scope="module")
 def qapp():
     from qtpy.QtWidgets import QApplication
@@ -10,7 +9,6 @@ def qapp():
     yield app
 
 
-@pytest.mark.slow
 @pytest.fixture(scope="module")
 def win(qapp):
     from cellpose.gui.gui3d import MainW_3d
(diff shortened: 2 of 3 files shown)


cellpose/gui/gui.py
```diff
@@ -177,7 +177,7 @@ class MainW(QMainWindow):
     def __init__(self, image=None, logger=None):
         super(MainW, self).__init__()
 
-        self.logger = logger
+        self.logger = logger if logger is not None else logging.getLogger(__name__)
```
mrariden, line 180: helpful, thanks

tests/test_gui.py
mrariden, line 1: These tests seem a bit fragile but we can keep them if they pass the GH workflow.

mrariden (commented):
See if the GH tests pass, and remove the tests if they dont

carsen-stringer:
Thanks for working on this! I used the original loading code for this instead of doing a new reimplementation

PR #1456, opened 2026-05-21
add weights and biases to cellpose training

jefromson:
I am looking to use weights and biases (WandB - ) logging during training to monitor progress and hoping this is useful to others. Please let me know if there is other specific information you would like included in this summary or if this is not wanted. Currently, training and test loss are the primary logged metrics which are generated during training. Parameters consumed here default to wandb environment variables.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -127,3 +127,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# Weights & Biases
+wandb/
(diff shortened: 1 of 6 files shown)


carsen-stringer:
Thanks for the implementation, unfortunately this is a fairly big change that we will not support as we do not use wandb (and won't be able to debug if it breaks). Fine-tuning cellpose models is fairly fast, and it's always better to be in the loop with hyperparameter searches, so we would be happy to support tensorboard, but not something hands off like this.

PR #1477, opened 2026-06-22
Fix PyQtGraph shape mismatch exception by passing levels synchronously

wong-ziyi:
### Description
When a user switches color views (e.g. from `rgb` to `gray`) while rendering a 3D image, the dimensionality of the array sent to `pyqtgraph` `ImageItem` can change. In `gui.py`'s `update_plot` function, `self.img.setImage` is called without the `levels` keyword argument, and then `self.img.setLevels(levels)` is called sequentially.

However, `pyqtgraph` asynchronously processes `setImage` by invoking `makeARGB` with the cached `self.levels` from the *previous* image. If the new image has a different shape but `pyqtgraph` tries to unpack it using the older multi-channel levels buffer, it triggers internal exceptions such as `ValueError: too many values to unpack (expected 2)` or `Exception: levels must have shape (data.shape[-1], 2)`, causing the GUI rendering to abort or crash entirely.

### Solution
This PR modifies `update_plot` in `cellpose/gui/gui.py` to synchronously supply the `levels` arguments directly into the `setImage` function parameters, which prevents asynchronous rendering mismatches. This matches best-practice guidelines for `pyqtgraph`.

### DIFF
--- a/cellpose/gui/gui.py
+++ b/cellpose/gui/gui.py
@@ -1482,22 +1482,21 @@ class MainW(QMainWindow):
             else: 
                 image = self.stack_filtered[self.currentZ]
             if self.color == 'rgb':
-                self.img.setImage(image, autoLevels=False, lut=None)
                 if self.nchan > 1:
                     levels = np.array([self.saturation[r][self.currentZ] for r in range(image.shape[-1])])
                 else: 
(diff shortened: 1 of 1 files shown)


wong-ziyi:

PR #1478, opened 2026-06-22
Fix Python 3.8 syntax error by removing 3.10 specific union type hint

wong-ziyi:
### Description
The recent updates to `gui.py` introduced a Python 3.10+ specific type hinting syntax (`str | int`) on the `color` property setter:
```python
    @color.setter
    def color(self, value: str|int):
```
This breaks compatibility for users running Python 3.8 or 3.9, resulting in a `TypeError: unsupported operand type(s) for |: 'type' and 'type'` upon starting the GUI.

### Solution
This PR simply removes the type hint (or could use `typing.Union[str, int]`) to restore compatibility with older supported Python versions.


wong-ziyi:

